PR #2, opened 2016-04-20
Travis config model

Soullivaneuh:
This travis file it's just a model at the moment.

Will work on a travis file generator later to dispatch on each repository.

### DIFF
--- a/tools/travis/.travis.yml
+++ b/tools/travis/.travis.yml
@@ -38,12 +38,12 @@ cache:
 
 before_install:
   - if [ "$TRAVIS_PHP_VERSION" != "hhvm" ]; then phpenv config-rm xdebug.ini; fi;
-
-install:
   - composer config -g github-oauth.github.com $GITHUB_OAUTH_TOKEN
   - composer global require phpunit/phpunit:@stable --no-update
-  - composer global update --prefer-dist --no-interaction
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; fi;
+
+install:
+  - composer global update --prefer-dist --no-interaction
   - composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
 
 before_script:
--- a/tools/travis/.travis.yml
+++ b/tools/travis/.travis.yml
@@ -43,7 +43,7 @@ install:
   - composer config -g github-oauth.github.com $GITHUB_OAUTH_TOKEN
   - composer global require phpunit/phpunit:@stable --no-update
   - composer global update --prefer-dist --no-interaction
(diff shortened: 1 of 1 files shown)


tools/travis/.travis.yml
```diff
@@ -0,0 +1,52 @@
...
+
+env:
+  global:
+    - PATH="$HOME/.composer/vendor/bin:$PATH"
+
+matrix:
+  fast_finish: true
+  include:
+    - php: 5.6
+      env: COMPOSER_FLAGS="--prefer-lowest"
```
core23: We should use php 5.3 here as it is the lowest supported php version.

tools/travis/.travis.yml
```diff
@@ -0,0 +1,52 @@
...
+    - $HOME/.composer/cache/files
+
+before_install:
+  - if [ "$TRAVIS_PHP_VERSION" != "hhvm" ]; then phpenv config-rm xdebug.ini; fi;
+
+install:
+  - composer config -g github-oauth.github.com $GITHUB_OAUTH_TOKEN
+  - composer global require phpunit/phpunit:@stable --no-update
+  - composer global update --prefer-dist --no-interaction
+  - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; PHPUNIT_OPTION="--testsuite symfony"; fi;
```
greg0ire: Please use the multiline syntax

tools/travis/.travis.yml
```diff
@@ -0,0 +1,52 @@
...
+  - composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
+
+before_script:
+  - if [ "$TRAVIS_PHP_VERSION" != "hhvm" ]; then echo "zend_extension=xdebug.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi;
```
greg0ire: Same here.

tools/travis/.travis.yml
Soullivaneuh: Indeed, bad copy paste.

tools/travis/.travis.yml
Soullivaneuh: Seriously? For one command line?

And what should be the proper format exactly?

greg0ire: Yeah seriously : 

```
- | if [ "$TRAVIS_PHP_VERSION" != "hhvm" ]; 
    then echo "zend_extension=xdebug.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; 
    fi;
```
is an example of how beautiful it could be come.

And I would argue that this is not just a command line because you make it a line. You're not at home in your terminal hacking something, you're scripting something. So use several lines, long options, and if it grows too much, move all that to a separate shell script that will have syntactic coloration and proper indention.

PR #3, opened 2016-04-20
EditorConfig model file

Soullivaneuh:
This file is supported by a lot of common editors.
It should be added on project repositories.

See: http://editorconfig.org/

### DIFF
--- a/tools/editorconfig/.editorconfig
+++ b/tools/editorconfig/.editorconfig
@@ -1,0 +1,20 @@
+root = true
+
+[*]
+end_of_line = lf
+indent_style = space
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{yml,twig,php}]
+indent_size = 4
+
+[.travis-ci.yml]
+indent_size = 2
+
+[composer.json]
+indent_size = 4
+
+[Makefile]
+indent_style = tab


tools/editorconfig/.editorconfig
```diff
@@ -0,0 +1,20 @@
...
+trim_trailing_whitespace = true
+
+[*.{yml,twig,php}]
+indent_size = 4
+
+[.travis-ci.yml]
+indent_size = 2
+
+[composer.json]
+indent_size = 4
```
ju1ius: Maybe you can just 4space all the things, and do per file(type) overrides:
```ini
[*]
indent_size = 4
[.travis-ci.yml]
indent_size = 2
```

tools/editorconfig/.editorconfig
```diff
@@ -0,0 +1,20 @@
...
+
+[*]
+end_of_line = lf
+indent_style = space
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{yml,twig,php}]
+indent_size = 4
```
ju1ius: Funny how javascript and css are missing.

greg0ire: Can you please take care of that part @ju1ius ?

Soullivaneuh: The reason is pretty simple: I get this one from one of my projects that don't have any css/js. :-)

tools/editorconfig/.editorconfig
Soullivaneuh: This is dangerous because if a new filetype is added that have, for example, 8 spaces of indent by convention, this will be overridden by editorconfig.

tools/editorconfig/.editorconfig
ju1ius: @Soullivaneuh @greg0ire The most used convention nowadays, is 2 spaces for js/css, but maybe being consistent with php files is better ?

Soullivaneuh: > but maybe being consistent with php files is better ?

Convention is convention. I don't see why we have to put 4 spaces everywhere because "PHP do it this way". :-)

PR #4, opened 2016-04-20
Some travis improvements

core23:
- Added deprecation helper for symfony 
- ~~require lowest stables releases~~

### DIFF
--- a/tools/travis/.travis.yml
+++ b/tools/travis/.travis.yml
@@ -12,6 +12,7 @@ php:
 env:
   global:
     - PATH="$HOME/.composer/vendor/bin:$PATH"
+    - SYMFONY_DEPRECATIONS_HELPER=weak
 
 matrix:
   fast_finish: true
@@ -26,9 +27,12 @@ matrix:
       env: SYMFONY_VERSION=3.0.*
     - php: 5.6
       env: SYMFONY_VERSION=3.1.*@dev
+    - php: 5.6
+      env: SYMFONY_DEPRECATIONS_HELPER=0
   allow_failures:
     - php: nightly
     - env: SYMFONY_VERSION=3.1.*@dev
+    - env: SYMFONY_DEPRECATIONS_HELPER=0
 
 sudo: false
 


core23:
I would like to add cs-fixer and sphinx build too (via Makefile), if there is nobody against this

greg0ire:
I am for Makefile because it is more powerful than composer scripts IMO (variables, syntactic coloration), but if people want to make a case for composer scripts, now is the time I think (looking at you, @Soullivaneuh ;) )

Soullivaneuh:
> I would like to add cs-fixer and sphinx build too (via Makefile), if there is nobody against this

I don't know how I'll manage it for now. Please don't add new things for the moment or just open issue.

PR on existing things are welcome.

tools/travis/.travis.yml
```diff
@@ -12,12 +12,13 @@ php:
...
   global:
     - PATH="$HOME/.composer/vendor/bin:$PATH"
+    - SYMFONY_DEPRECATIONS_HELPER=weak
 
 matrix:
   fast_finish: true
   include:
     - php: 5.3
-      env: COMPOSER_FLAGS="--prefer-lowest"
+      env: COMPOSER_FLAGS="--prefer-lowest --prefer-stable"
```
Soullivaneuh: Are you sure it changes something?

Lowest deps will get stable release if exists. Do you have a concrete case?

Soullivaneuh:
@core23 Could you please add a case with `SYMFONY_DEPRECATIONS_HELPER=0` on `allow_failures` section?

http://symfony.com/doc/2.8/components/phpunit_bridge.html#making-tests-fail

tools/travis/.travis.yml
core23: We need this, if we have something like this in the **composer.json** file

``
  "minimum-stability": "dev",
``

greg0ire: I think we should avoid `minimum-stability` at all costs and specify dev dependencies explicitely instead. I have come to believe Composer is less efficient with it.

core23: It doesn't bother the normal Travis build, but it would help in this concrete case.

Soullivaneuh: With the following `composer.json`:

```json
{
    "name": "root/composer-check",
    "require": {
        "symfony/http-kernel": "^2.3"
    },
    "minimum-stability": "dev",
    "prefer-stable": true
}
```

And the following command:

```bash
composer up --prefer-lowest
```

I got:

```bash
composer info 
psr/log                  1.0.0  Common interface for logging libraries
symfony/debug            v2.3.0 Symfony Debug Component
symfony/event-dispatcher v2.1.0 Symfony EventDispatcher Component
symfony/http-foundation  v2.2.0 Symfony HttpFoundation Component
symfony/http-kernel      v2.3.0 Symfony HttpKernel Component
```

Soullivaneuh: I also tried without `"prefer-stable": true` on my `composer.json` file. Same results.

@core23 Could you please give me a concrete working case of what you are saying?

core23:
Removed the prefer-stable flag. Maybe we could add it later, when we got a practial usecase.

tools/travis/.travis.yml
```diff
@@ -26,9 +27,11 @@ matrix:
       env: SYMFONY_VERSION=3.0.*
     - php: 5.6
       env: SYMFONY_VERSION=3.1.*@dev
+    - env: SYMFONY_DEPRECATIONS_HELPER
```
Soullivaneuh: I think you have to specify a PHP version. Otherwise this env will create a job for each version.

Soullivaneuh:
Thank you @core23

PR #9, opened 2016-04-20
PHP-CS-Fixer + StyleCI config

Soullivaneuh:


tools/php-cs/.styleci.yml
```diff
@@ -0,0 +1,23 @@
+# DO NOT EDIT THIS FILE!
+#
+# It's auto-generated by sonata-project/dev-kit package.
+#
+# Package `sllh/php-cs-fixer-styleci-bridge` is required to get it working.
+
+preset: symfony
+
+enabled:
+  - align_double_arrow
```
ju1ius: I think a lot of people do not like this one...

Soullivaneuh: I understand but it was a @rande request. :stuck_out_tongue: 

For now I'll keep as is to get it working. Feel free to open a PR once this one merged to start a discussion about it. :+1:

greg0ire: See #11

ju1ius: I think the closer it stays from PSR-2 + Symfony cs rules, the less friction for everybody.

greg0ire: @ju1ius discussion is in #11

PR #11, opened 2016-04-22
Force unaligning equal signs and arrows

greg0ire:
Alignment can seem like a good idea at first, but it makes diff harder
to read, and bigger, thus increasing the risk of conflicts. It also adds
a load on the developer because it requires maintenance when they add or
remove the longest key of the array / variable of the assignment list.

### DIFF
--- a/project/.styleci.yml
+++ b/project/.styleci.yml
@@ -7,7 +7,6 @@
 preset: symfony
 
 enabled:
-  - align_double_arrow
   - newline_after_open_tag
   - ordered_use
   - long_array_syntax
@@ -16,10 +15,6 @@ enabled:
 #  - strict_param
 #  - php_unit_construct
 #  - php_unit_strict
-
-disabled:
-  - unalign_double_arrow
-  - unalign_equals
 
 finder:
   exclude:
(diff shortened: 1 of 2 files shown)


Soullivaneuh:
I personally agree with you. I originally added this rule by asking of @rande :https://github.com/sonata-project/SonataAdminBundle/pull/3038#discussion_r31956865

But your suggestion make sens. @rande what do you think?

rande:
@Soullivaneuh @greg0ire is the issue really happen a lot vs reading the code ?

greg0ire:
> @Soullivaneuh @greg0ire is the issue really happen a lot vs reading the code ?

In my experience it does, but I don't think the most affected people are reviewers, but contributors : they have to maintain the alignment, and it can be very cumbersome, especially if you have nested arrays.

Soullivaneuh:
Agree with @greg0ire. And if a user add a longer key, this is the whole array that would be on the diff.

Harder to read, and more risky about merge conflicts between PRs.

rande:
I just want to use the same convention of symfony which seems to respect this convention: "Force unaligning equal signs and arrows".

so 👍

Soullivaneuh:
Ok then. I'll merge this one once StyleCI will be enabled on all project.

StyleCI have a system of auto PR creation, it will be used. :+1:

Soullivaneuh:
@greg0ire Could you please rebase?

greg0ire:
@Soullivaneuh : done

PR #15, opened 2016-04-22
adds indent size settings for js, json, scss & css

ju1ius:
A popular convention is to use 2 spaces for indenting js & css files.
One could also choose 4 spaces, to favour consistency with other file types among the project.
I do not have any kind of preference on this really, so just react with:
* :+1: if you're ok with 2 spaces
* :-1: if you prefer 4

### DIFF
--- a/tools/editorconfig/.editorconfig
+++ b/tools/editorconfig/.editorconfig
@@ -10,6 +10,9 @@ trim_trailing_whitespace = true
 [*.{yml,twig,php}]
 indent_size = 4
 
+[*.{js,json,scss,css}]
+indent_size = 2
+
 [.travis-ci.yml]
 indent_size = 2
 


tools/editorconfig/.editorconfig
```diff
@@ -10,6 +10,9 @@ trim_trailing_whitespace = true
 [*.{yml,twig,php}]
 indent_size = 4
 
+[*.{js,css}]
```
Soullivaneuh: Do we have another asset extension like `.less`, `.scss`, `.coffee` on sonata projects?

ju1ius:
```sh
cd sonata/sandbox/vendor/sonata-project
for ext in {scss,sass,less,coffee}; do
  count=$(find . -path '*/vendor/*' -o -path '*/node_modules/*' -prune -o -name "*.$ext" -print | wc -l)
  echo "$ext: $count"
done
```
```
scss: 4
sass: 0
less: 0
coffee: 0
```
So yes, scss files in page-bundle

Soullivaneuh:
Thank you @ju1ius

PR #16, opened 2016-04-22
adds indent size setting for package.json

ju1ius:
Use npm's default indent size, so that people don't mess with it accidentally
when editing the file by hand.

### DIFF
--- a/tools/editorconfig/.editorconfig
+++ b/tools/editorconfig/.editorconfig
@@ -10,7 +10,7 @@ trim_trailing_whitespace = true
 [*.{yml,twig,php}]
 indent_size = 4
 
-[.travis-ci.yml]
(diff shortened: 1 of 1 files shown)


tools/editorconfig/.editorconfig
```diff
@@ -10,7 +10,7 @@ trim_trailing_whitespace = true
 [*.{yml,twig,php}]
 indent_size = 4
 
-[.travis-ci.yml]
+[{package.json,.travis-ci.yml}]
```
Soullivaneuh: AFAIK, `.json` are indented by two spaces by default.

Soullivaneuh: And BTW, another `.json` files should be two spaced too. `composer.json` is the **only** exception.

PR #29, opened 2016-05-01
Label update automation

Soullivaneuh:
The goal of this PR is to update all labels of all sonata project.

For the moment, it just output what it should do.

@sonata-project/contributors Please take profits to discuss which label should be added on the list and why (just for label, not color).

This should be done now because all other not wanted labels **will be destroyed**.

You can see the output on Travis jobs.

- [x] Make update real (test only on `cache` and `exporter` for now)
- [x] Only show updated/deleted label
- [x] Success message at end

### DIFF
--- a/src/Config/Configuration.php
+++ b/src/Config/Configuration.php
@@ -1,0 +1,53 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DevKit\Config;
+
+use Symfony\Component\Config\Definition\Builder\TreeBuilder;
+use Symfony\Component\Config\Definition\ConfigurationInterface;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+class Configuration implements ConfigurationInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder();
+        $rootNode = $treeBuilder->root('sonata');
+
+        $rootNode
+            ->children()
+                ->arrayNode('labels')
+                    ->isRequired()
+                    ->requiresAtLeastOneElement()
+                    ->useAttributeAsKey('name')
+                    ->prototype('array')
+                        ->children()
+                            ->scalarNode('color')->isRequired()->cannotBeEmpty()->end()
+                        ->end()
+                    ->end()
+                ->end()
+                ->arrayNode('projects')
+                    ->isRequired()
+                    ->requiresAtLeastOneElement()
+                    ->normalizeKeys(false)
+                    ->prototype('scalar')->end()
+                ->end()
+            ->end()
+        ;
+
+        return $treeBuilder;
+    }
+}
--- a/src/Console/Command/DispatchCommand.php
+++ b/src/Console/Command/DispatchCommand.php
@@ -77,11 +77,7 @@ class DispatchCommand extends Command
 
         $this->packagistClient = new \Packagist\Api\Client();
 
-        $this->githubClient = new \Github\Client(
-            new \Github\HttpClient\CachedHttpClient(array(
-                'cache_dir' => sys_get_temp_dir().'/github-api-cache',
-            ))
-        );
+        $this->githubClient = new \Github\Client();
         if (getenv('GITHUB_OAUTH_TOKEN')) {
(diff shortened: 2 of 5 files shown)


Th3Mouk:
@Soullivaneuh what do think about `help needed`. I remember when some tests not passed and i was blocked during lot of week, the community can't distinguish.

Can you also explained `vendor` and `rtm` please ?

I don't know if it's the same think of `documentation`, for PR on the doc.

Soullivaneuh:
> what do think about `help needed`. I remember when some tests not passed and i was blocked during lot of week, the community can't distinguish.

:+1: 

> Can you also explained `vendor` and `rtm` please ?

`vendor`: When the issue/PR is related to the vendor. Example: One issue appear on SonataUserBundle but it's caused by a bug introduced on a new version of FOSUserBundle.

RTM: Ready to Merge. This should be used for example when a PR is OK but CI is not finished. In this case, put this label so you can remember to merge later if CI passes.

> I don't know if it's the same think of documentation, for PR on the doc.

:+1: for that but I think I would name it `doc` only because it's already used by many projects of sonata.

core23:
- docs
- review required - someone else should verify the PR
- test missing (?)

greg0ire:
I'd like a label for cross-validation : I'm reluctant to merge things myself if I'm the only reviewer. `review_required` does not really mean that, does it. I would like a label to mean "ok for me, if it's ok for someone else then merge".

Soullivaneuh:
> test missing (?)

Would not be enough to tell user to add test and add "Pending author" label?

src/Config/Configuration.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+class Configuration implements ConfigurationInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder();
+        $rootNode = $treeBuilder->root('sonata');
```
OskarStark: should be `sonata_dev_kit` IMO

OskarStark:
what about `WIP` ?

src/Console/Command/DispatchCommand.php
```diff
@@ -45,13 +64,94 @@ protected function configure()
...
+
+        $configs = Yaml::parse(file_get_contents(__DIR__.'/../../../.sonata-project.yml'));
+        $processor = new Processor();
+        $this->configs = $processor->processConfiguration(new Configuration(), array('sonata' => $configs));
+
+        $this->packagistClient = new \Packagist\Api\Client();
+
+        $this->githubClient = new \Github\Client(
+            new \Github\HttpClient\CachedHttpClient(array(
+                'cache_dir' => sys_get_temp_dir().'/github-api-cache',
```
greg0ire: A good practice I found by suffering from crowded `/tmp/` directories on my servers is to also include the name of the project that generates stuff and sometimes does not clean afterwards. So using `/sonata-dev-kit/github-api-cache` would be better IMO

src/Console/Command/DispatchCommand.php
```diff
@@ -45,13 +64,94 @@ protected function configure()
...
+        $missingLabels = $configuredLabels;
+
+        $headers = array('Name', 'Actual color', 'Needed Color', 'State');
+        $rows = array();
+
+        foreach ($this->githubClient->repository()->labels()->all('sonata-project', $repositoryName) as $label) {
+            $name = $label['name'];
+            $color = $label['color'];
+
+            $shouldExists = array_key_exists($name, $configuredLabels);
```
greg0ire: `shouldExist`

src/Console/Command/DispatchCommand.php
Soullivaneuh: Not relevant anymore. Cache causes trouble on execution, removed it for now.

src/Config/Configuration.php
Soullivaneuh: Not a big deal, this is not used on the yaml file.

Soullivaneuh:
> what about `WIP` ?

@OskarStark You mean this one? https://github.com/sonata-project/SonataAdminBundle/labels/in%20progress

Soullivaneuh:
I am not sure for the "In progress" one because the contributor can't set it himself.

Adding a `[WIP]` prefix on title would be simpler IMHO.

OskarStark:
> @OskarStark You mean this one? https://github.com/sonata-project/SonataAdminBundle/labels/in%20progress

yes, `in progress` is fine, too

OskarStark:
What do you think of "mapping" existing labels instead of only deleting them?
Could it be useful?

Soullivaneuh:
> What do you think of "mapping" existing labels instead of only deleting them?

This is an open door of lot of duplicate and/or unused labels.

If we do that, so would be better to do nothing.

With this we will have consistency. If a label should be added, we can open PR later to discuss it.

Th3Mouk:
no tags when help is needed ?

Soullivaneuh:
Oh sorry @Th3Mouk I forgot it!

PR #31, opened 2016-05-02
Dispatch files

Soullivaneuh:
- [x] Make diff from files provided on config
- [x] Apply commit if diff on branch
- [x] Make placeholder to replace like `{{ stable_branch }}`, `{{ project_name }}`...


.sonata-project.yml
```diff
@@ -24,6 +24,10 @@ labels:
   - name: help wanted
     color: '0e8a16'
 
+files:
+  - doc/CONTRIBUTING.md:CONTRIBUTING.md
+  - tools/editorconfig/.editorconfig:.editorconfig
```
greg0ire: Maybe we should keep it simple and have only one tree, matching exactly the destination tree. That way we may simply use a recursive copy.

Soullivaneuh: Yes maybe. Have to test, didn't decide yet.

We also have more complicated file like `.travis.yml` that will needs a twig template...

greg0ire: Puppet separates into a `file` folder and a `templates` directory, but I guess having different extensions for files in the same directory would be enough…

Soullivaneuh:
@greg0ire I updated the code to manage directories. What do you think?

greg0ire:
IMO, if you still need the `files` config entry, there is a problem. One should be able to sync a new file without having to configure anything. In [git template](https://github.com/greg0ire/git_template), I segregated files that are supposed to be rsynced in a separate directory (`template`).

Soullivaneuh:
@greg0ire like this so?

greg0ire:
Exactly :+1:

src/Console/Command/DispatchCommand.php
```diff
@@ -12,28 +12,37 @@
...
 use Symfony\Component\Yaml\Yaml;
 
 /**
  * @author Sullivan Senechal <soullivaneuh@gmail.com>
  */
 class DispatchCommand extends Command
 {
+    const GITHUB_GROUP = 'sonata-project';
+    const GITHUB_USER = 'SonataCI';
+    const GITHUB_EMAIL = 'ci@sonata-project.org';
```
Soullivaneuh: @rande the PR is ready. All I need before merge is the real email of `SonataCI` user.

Could you please give me this one?

src/Console/Command/DispatchCommand.php
```diff
@@ -197,4 +236,116 @@ private function updateLabels($repositoryName)
...
+            ->config('user.name', static::GITHUB_USER)
+            ->config('user.email', static::GITHUB_EMAIL)
+        ;
+
+        foreach ($projectConfig['branches'] as $branch => $branchConfig) {
+            $this->io->section('Files for '.$branch);
+
+            $git
+                ->reset(array('hard' => true))
+                ->checkout('-b', $branch, '-t', 'origin/'.$branch)
```
greg0ire: Using long options would be more explicit (`--track` instead of `-t`)

PR #80, opened 2016-05-18
Manage RTM and 'review required' labels.

Soullivaneuh:
- If a PR is opened or updated, 'review required' is set.
- If a PR is updated and 'RTM' is set, it is removed.

This is a pre-sequel of PR merging automation.

### DIFF
--- a/src/GithubHookProcessor.php
+++ b/src/GithubHookProcessor.php
@@ -55,11 +55,67 @@ final class GithubHookProcessor
         $commentAuthorId = $payload['comment']['user']['id'];
 
         if ($commentAuthorId === $issueAuthorId) {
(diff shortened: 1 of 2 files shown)


src/GithubHookProcessor.php
```diff
@@ -55,11 +55,66 @@ public function processPendingAuthor($eventName, array $payload)
...
+     * @param string $repoName
+     * @param int    $issueId
+     * @param string $label
+     */
+    private function addIssueLabel($repoUser, $repoName, $issueId, $label)
+    {
+        $isSet = false;
+        foreach ($this->githubClient->issues()->labels()->all($repoUser, $repoName, $issueId) as $labelInfo) {
+            if ($label === $labelInfo['name']) {
+                $isSet = true;
```
greg0ire: Use `return` here and `isSet` will become useless.

PR #87, opened 2016-05-19
Auto message for conflicted PR

### LINKED ISSUE #82: Conflict resolve need automatic message
SonataCI could help us to tell PR author to rebase a conflicted one.

- Say a "Could you please rebase and fix merge conflicts?" message
- Add a `pending author` label

Soullivaneuh:
Closes #82.

- [x] "Please rebase" message
- [x] "pending author" label

### DIFF
--- a/src/Console/Command/MergeConflictsCommand.php
+++ b/src/Console/Command/MergeConflictsCommand.php
@@ -1,0 +1,93 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DevKit\Console\Command;
+
+use Github\Exception\ExceptionInterface;
+use Packagist\Api\Result\Package;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+final class MergeConflictsCommand extends AbstractNeedApplyCommand
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        parent::configure();
+
+        $this
+            ->setName('merge-conflicts')
+            ->setDescription('Comments non-mergeable pull requests, asking the author to solve conflicts.')
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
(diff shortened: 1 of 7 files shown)


src/Console/Command/MergeConflictsCommand.php
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        parent::configure();
+
+        $this
+            ->setName('merge-conflicts')
+            ->setDescription('Put a conflicts resolve asking on not mergeable Pull Requests.')
```
greg0ire: Asking is not a noun. Maybe

> Comments non-mergeable pull requests, asking the author to solve conflicts.

src/Console/Command/MergeConflictsCommand.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        foreach ($this->configs['projects'] as $name => $projectConfig) {
+            try {
+                $package = $this->packagistClient->get(static::PACKAGIST_GROUP.'/'.$name);
+                $this->io->title($package->getName());
+                $this->checkPullRequests($package);
+            } catch (ExceptionInterface $e) {
+                $this->io->error('Failed with message: '.$e->getMessage());
```
greg0ire: Not sure dropping the exception stack trace is a good thing… isn't their a dedicated method to log exceptions in the `$this->io` object?

Soullivaneuh: The IO indeed log the thrown exception but the comment is stopped just after that.

The goal is to continue on next project if we have fails on the current one.

src/Console/Command/MergeConflictsCommand.php
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private function checkPullRequests(Package $package)
+    {
+        $repositoryName = $this->getRepositoryName($package);
+
+        foreach ($this->githubClient->pullRequests()->all(static::GITHUB_GROUP, $repositoryName) as $pullRequest) {
+            $number = $pullRequest['number'];
+            $pullRequest = $this->githubClient->pullRequests()->show(static::GITHUB_GROUP, $repositoryName, $number);
+            if (false === $pullRequest['mergeable']) {
```
greg0ire: Why make a strict check here? If `$pullRequest['mergeable'] === 0` instead of `false`, this piece of code will do not anything to report it.

Soullivaneuh: Because of the `null` value.

> The value of the mergeable attribute can be true, false, or null. If the value is null, this means that the mergeability hasn't been computed yet, and a background job was started to compute it. Give the job a few moments to complete, and then submit the request again. When the job is complete, the response will include a non-null value for the mergeable attribute.

From: https://developer.github.com/v3/pulls/#get-a-single-pull-request

PR #91, opened 2016-05-20
Use shell scripts for Travis

### LINKED ISSUE #59: The travis file is getting ugly
Some thoughts:

- YML is not the best language to write commands, shell is (because of syntactic coloration and things like [ShellCheck](https://github.com/koalaman/shellcheck)).
- Only Travis can run the commands, if you want to reproduce a failed build locally, you have to do some copy pasting.
- If me move to another CI solution, some things will change but some things should not.

In the end it would be great to do something like this :

```yml
before_install:
   - source scripts/ci.sh
   - before_install # call a function defined in ci.sh
```

Soullivaneuh:
Fixes #59.

The goal is to have a nicer view of the configured scripts.

Travis build: https://travis-ci.org/sonata-project/exporter/builds/131681135

### DIFF
--- a/project/.travis/install_test.sh
+++ b/project/.travis/install_test.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env sh
+set -ev
+
+mkdir --parents "${HOME}/bin"
+
+# PHPUnit install
+if [ ${TRAVIS_PHP_VERSION} '<' '5.6' ]; then
+    PHPUNIT_PHAR=phpunit-old.phar
+else
+    PHPUNIT_PHAR=phpunit.phar
+fi
+wget "https://phar.phpunit.de/${PHPUNIT_PHAR}" --output-document="${HOME}/bin/phpunit"
+chmod u+x "${HOME}/bin/phpunit"
+
+composer update --prefer-dist --no-interaction --prefer-stable ${COMPOSER_FLAGS}
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -53,25 +53,12 @@ matrix:
 {% endfor %}
 
 before_install:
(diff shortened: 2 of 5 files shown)


project/.travis/before_install_docs.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env sh
+set -ev
```
greg0ire: Maybe check if the file exists before executing it instead?

Soullivaneuh: I prefer keep this file even if is empty.

With this, people know that those files are managed by dev-kit and not by the project.

project/.travis/install_test.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env sh
+set -ev
+
+mkdir --parent "${HOME}/bin"
```
greg0ire: `parent` => `parents`

project/.travis/install_test.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# PHPUnit install
+if [ ${TRAVIS_PHP_VERSION} '<' '5.6' ]; then
+    wget https://phar.phpunit.de/phpunit-old.phar --output-document="${HOME}/bin/phpunit"
+else
+    wget https://phar.phpunit.de/phpunit.phar --output-document="${HOME}/bin/phpunit"
+fi
```
greg0ire:
```sh
if [ ${TRAVIS_PHP_VERSION} '<' '5.6' ]; then
    PHPUNIT_PHAR=phpunit-old.phar
else
    PHPUNIT_PHAR=phpunit.phar
fi
wget "https://phar.phpunit.de/${PHPUNIT_PHAR}" --output-document="${HOME}/bin/phpunit"
```

Soullivaneuh:
@greg0ire Are we ok now ?

greg0ire:
Perfect :+1:

PR #100, opened 2016-05-23
Automatic merging

### LINKED ISSUE #96: Branch merging automation
Since we have a homogeneous branch management, maybe we can introduce branch auto merging.

A cron task could run every night to merge all stable branches of each repo into the master one.

Of course, branches with conflicts will be ignored.

This will avoid long branch difference if we forgot to merge them from a while...

In addition, a web page could be added later, showing branches that need manual merge.

@sonata-project/contributors What do you think?

Soullivaneuh:
Closes #96.

We need https://github.com/KnpLabs/php-github-api/pull/379 to be merged and on a tag to get this working.

After merge this PR, a cron task have to be setup to run every night.

### DIFF
--- a/src/Console/Command/AutoMergeCommand.php
+++ b/src/Console/Command/AutoMergeCommand.php
@@ -1,0 +1,97 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\DevKit\Console\Command;
+
+use Github\Exception\ExceptionInterface;
+use Github\Exception\RuntimeException;
+use Packagist\Api\Result\Package;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+final class AutoMergeCommand extends AbstractNeedApplyCommand
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        parent::configure();
+
+        $this
+            ->setName('auto-merge')
+            ->setDescription('Merges branches of repositories if there is no conflict.')
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        foreach ($this->configs['projects'] as $name => $projectConfig) {
+            try {
+                $package = $this->packagistClient->get(static::PACKAGIST_GROUP.'/'.$name);
+                $this->io->title($package->getName());
+                $this->mergeBranches($package, $projectConfig);
+            } catch (ExceptionInterface $e) {
+                $this->io->error('Failed with message: '.$e->getMessage());
+            }
+        }
+
+        return 0;
+    }
+
+    private function mergeBranches(Package $package, array $projectConfig)
+    {
+        if (!$this->apply || !array_key_exists('branches', $projectConfig)) {
+            return;
+        }
+
+        $repositoryName = $this->getRepositoryName($package);
+        $branches = array_reverse(array_keys($projectConfig['branches']));
+
+        // Merge the oldest branch into the next newest, and so on.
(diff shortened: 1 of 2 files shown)


Soullivaneuh:
Added a manual message waiting https://github.com/KnpLabs/php-github-api/pull/379.

src/Console/Command/AutoMergeCommand.php
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        parent::configure();
+
+        $this
+            ->setName('auto-merge')
+            ->setDescription('Merges branches of repositories if not conflict.')
```
greg0ire: "if they don't conflict" or "if there is no conflict"

src/Console/Command/AutoMergeCommand.php
```diff
@@ -0,0 +1,97 @@
...
+            try {
+                $package = $this->packagistClient->get(static::PACKAGIST_GROUP.'/'.$name);
+                $this->io->title($package->getName());
+                $this->mergeBranches($package, $projectConfig);
+            } catch (ExceptionInterface $e) {
+                $this->io->error('Failed with message: '.$e->getMessage());
+            }
+        }
+
+        return 0;
```
greg0ire: Maybe return something else if there has been an exception ?

Soullivaneuh: Not really. The fail is for one project, the command still continue.

src/Console/Command/AutoMergeCommand.php
```diff
@@ -0,0 +1,97 @@
...
+        if (!$this->apply || !array_key_exists('branches', $projectConfig)) {
+            return;
+        }
+
+        $repositoryName = $this->getRepositoryName($package);
+        $branches = array_reverse(array_keys($projectConfig['branches']));
+
+        while (($head = current($branches))) {
+            $base = next($branches);
+            if (false === $base) {
```
greg0ire: I don't get this part… why the reverse, why not use foreach, and why check for `false`?

src/Console/Command/AutoMergeCommand.php
```diff
@@ -0,0 +1,97 @@
...
+                if (is_array($response) && array_key_exists('sha', $response)) {
+                    $this->io->success('Merged '.$head.' into '.$base);
+                } else {
+                    $this->io->comment('Nothing to merge on '.$base);
+                }
+            } catch (RuntimeException $e) {
+                if (409 === $e->getCode()) {
+                    $this->io->warning('Merging of '.$head.' into '.$base.' contains conflicts. Skipped.');
+                    continue;
+                } else {
```
greg0ire: No need for the else statement I think (because there is a `continue`)

src/Console/Command/AutoMergeCommand.php
Soullivaneuh: > why the reverse

Because we merge the oldest branch to the newest.

> why not use foreach

Proper. We avoid to have to create an index variable to check the next one.

> and why check for false

`next` return `false` if there is nothing after the current element.

See: http://php.net/manual/fr/function.next.php

src/Console/Command/AutoMergeCommand.php
greg0ire: Why return anything in that case? Isn't `0` the default value ?

src/Console/Command/AutoMergeCommand.php
greg0ire: Could you add a comment "merge the oldest branch into the newest, and so on" ?

src/Console/Command/AutoMergeCommand.php
Soullivaneuh: `0|void|null`.


Yes I know. I prerfer return a number like shell script does on succes/failure.

src/Console/Command/AutoMergeCommand.php
Soullivaneuh: Done.

src/Console/Command/AutoMergeCommand.php
greg0ire: Do you also `return 200` in your http controllers :trollface: ?

Soullivaneuh: > Do you also `return 200` in your http controllers :trollface: ?

Controllers return `200` by default. :trollface:

PR #102, opened 2016-05-26
added CREDITS file

OskarStark:
Refs https://github.com/sonata-project/SonataAdminBundle/pull/3880

removed `famfam icons`, added `font-awesome icons` instead


Soullivaneuh:
I think you can add AdminLTE.

OskarStark:
I added AdminLTE, shall we really add the project which are using it?

Soullivaneuh:
> I added AdminLTE, shall we really add the project which are using it?

Maybe something like "For admin it components user interface"?

OskarStark:
ping @Soullivaneuh

project/CREDITS.md
```diff
@@ -0,0 +1,47 @@
...
+
+*jQuery is a fast, small, and feature-rich JavaScript library. It makes
+things like HTML document traversal and manipulation, event handling,
+animation, and Ajax much simpler with an easy-to-use API that works across
+a multitude of browsers. With a combination of versatility and extensibility,
+jQuery has changed the way that millions of people write JavaScript.*
+
+**License** https://github.com/jquery/jquery/blob/master/LICENSE.txt
+
+### Sonata Usecase
```
Soullivaneuh: Not sure of a title for that. Maybe same level as License?

OskarStark: will change that

core23:
Sure we need this on ALL bundles?

Soullivaneuh:
> Sure we need this on ALL bundles?

@core23 can you please elaborate?

core23:
I am sure if all bundles need a Credit for AdminLTE, jQuery, Font Awesome, ... 

E.g. why should we credit frontend librarys on the [exporter bundle](https://github.com/sonata-project/exporter)? There is no reference to any of them.

OskarStark:
> E.g. why should we credit frontend librarys on the exporter bundle? There is no reference to any of them.

for me its a matter of consistency, because the project itself says "thank you" and its much easier to maintain with the dev kit

core23:
I'am not against this PR in general. I'll just want to show up that we don't need this on all bundles.

OskarStark:
> I'am not against this PR in general. I'll just want to show up that we don't need this on all bundles.

You are true, but i think this file is not so important to build a logic, WHERE it should be updated by the dev-kit

ping @greg0ire

Soullivaneuh:
I would like @rande opinion about this file.

rande:
@core23 has a point. Also, credit should also refer to the PHP code used: bundles and libs. I think this reference works need to go on a dedicated page on the sonata-project.org website. Something similar to https://sonata-project.org/contributors.

greg0ire:
That would be a far better solution indeed IMO (the page is much more likely to be read than the file).

Soullivaneuh:
@rande Can the website be open-sourced? Or maybe a part? It would be great to permit contributions on it.

PR #111, opened 2016-05-28
Recommend against editing the change log file

greg0ire:
This should prevent many conflicts between PRs.

### DIFF
--- a/project/CONTRIBUTING.md
+++ b/project/CONTRIBUTING.md
@@ -84,14 +84,17 @@ on your PR explaining why you feel it is the case.
 
 #### The Change log
 
-On each PR, the `CHANGELOG.md` file **has to be updated**.
+For each PR, a change log must be provided.
 
-There are few cases where the `CHANGELOG.md` file should not be touched:
+There are few cases where no change log is necessary:
 
 * When you fix a bug on an unreleased feature.
 * When your PR concerns only the documentation (fix or improvement).
 
-The change log note has to be put below the `## [Unreleased]` section.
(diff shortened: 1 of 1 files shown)


project/CONTRIBUTING.md
```diff
@@ -84,14 +84,17 @@ on your PR explaining why you feel it is the case.
 
 #### The Change log
 
-On each PR, the `CHANGELOG.md` file **has to be updated**.
+For each PR, a change log SHOULD be provided.
 
-There are few cases where the `CHANGELOG.md` file should not be touched:
+There are few cases where no change log is necessary
```
Soullivaneuh: Missing `:` at the end.

project/CONTRIBUTING.md
```diff
@@ -84,14 +84,17 @@ on your PR explaining why you feel it is the case.
 
 #### The Change log
 
-On each PR, the `CHANGELOG.md` file **has to be updated**.
+For each PR, a change log SHOULD be provided.
```
Soullivaneuh: I would prefer `MUST`, even if it does not for few cases.

greg0ire: Well then I will use "must" instead of `MUST`

greg0ire: Otherwise people will think we refer to RFC 2119

Soullivaneuh: :+1:

Soullivaneuh:
Thank you! :+1:

PR #117, opened 2016-05-30
Dispatch on needed branches only

### LINKED ISSUE #115: Dispatch: Do PR only on stable first, then on master
The goal is to reduce the Travis load when the Dispatcher is running.

Most of the case, PR on stable and master branch are the same.

The proposed logic:

1. Check if some diff should be applied on stable branch
2. If yes, make a PR
3. Check if a PR is present for stable branch
4. If not, check if the stable branch is fully merged on master
5. If not and diff are present, make a PR against master

Soullivaneuh:
Closes #115.

### DIFF
--- a/src/Console/Command/DispatchCommand.php
+++ b/src/Console/Command/DispatchCommand.php
@@ -207,26 +207,50 @@ final class DispatchCommand extends AbstractNeedApplyCommand
             ->config('user.email', static::GITHUB_EMAIL)
         ;
 
-        foreach ($projectConfig['branches'] as $branch => $branchConfig) {
-            $this->io->section('Files for '.$branch);
+        $branches = array_reverse($projectConfig['branches']);
+
+        $previousBranch = null;
+        $previousDevKit = null;
+        while (($branchConfig = current($branches))) {
+            // We have to fetch all branches on each step in case a PR is submitted.
+            $remoteBranches = array_map(function ($branch) {
+                return $branch['name'];
+            }, $this->githubClient->repos()->branches(static::GITHUB_GROUP, $repositoryName));
+
+            $currentBranch = key($branches);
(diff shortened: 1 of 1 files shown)


src/Console/Command/DispatchCommand.php
```diff
@@ -207,52 +207,80 @@ private function dispatchFiles(Package $package)
...
+        $previousDevKit = null;
+        while (($branchConfig = current($branches))) {
+            // We have to fetch all branches on each step in case a PR is submitted.
+            $remoteBranches = array_map(function ($branch) {
+                return $branch['name'];
+            }, $this->githubClient->repos()->branches(static::GITHUB_GROUP, $repositoryName));
+
+            $currentBranch = key($branches);
+            $currentDevKit = $currentBranch.'-dev-kit';
+            next($branches);
```
greg0ire: Putting this right in the middle of the `while` statement feels a bit weird… why not line 215 ?

src/Console/Command/DispatchCommand.php
```diff
@@ -207,52 +207,80 @@ private function dispatchFiles(Package $package)
...
+
+            // A PR is already here for previous branch, do nothing on the current one.
+            if (in_array($previousDevKit, $remoteBranches, true)) {
+                continue;
+            }
+            // If the previous branch is not merged to the current one, do nothing.
```
greg0ire: "merged to" => "merged into"

src/Console/Command/DispatchCommand.php
Soullivaneuh: Because `key($branches);` would not give the same result after the `next`.

greg0ire: Right

greg0ire:
LGTM

PR #127, opened 2016-06-01
Add paragraph about deprecation_warnings

greg0ire:


project/CONTRIBUTING.md
```diff
@@ -186,6 +186,25 @@ interface BarInterface
...
 ```
 
+In some cases, you will have the possibility to warn the user that things will change,
+and recommend a new way of doing things. You can do so by triggering the dedicated kind of error, like this:
+
+```php
+<?php
+if (/* some condition showing the user is using the legacy way */) {
+    @trigger_error(
+        'Some message about this usage being deprecated in favor of something else'.
```
Soullivaneuh: * `.` -> `,`
* Maybe we can put proper examples like `'The '.__METHOD__.' method is deprecated since 3.x, to be removed in 4.0'`.
Or maybe links to good examples present on our code?

project/CONTRIBUTING.md
```diff
@@ -186,6 +186,25 @@ interface BarInterface
...
+        E_USER_DEPRECATED
+    );
+} else {
+    // new way of doing things
+}
+```
+
+In that case, unit tests might show your deprecation notice. You must mark such tests
+with the `@group legacy` annotation, and if need be,
```
Soullivaneuh: > with the `@group legacy` annotation,

-> On the previsous line.

> and if need be,

-> On the second.

project/CONTRIBUTING.md
greg0ire: >  links to good examples present on our code

Nah, don't make people click : it's far and you risk a 404. I'll give a proper example instead.

greg0ire:
@Soullivaneuh fixed

project/CONTRIBUTING.md
```diff
@@ -186,6 +186,24 @@ interface BarInterface
...
 ```
 
+In some cases, you will have the possibility to warn the user that things will change,
+and recommend a new way of doing things. You can do so by triggering the dedicated kind of error, like this:
+
+```php
+<?php
+if (/* some condition showing the user is using the legacy way */) {
+    @trigger_error(
+        'The '.__METHOD.' method is deprecated since 42.x, to be removed in 43.0',
```
Soullivaneuh: `method is deprecated since version 3.1 and will be removed in 4.0.`

From Symfony code sample.

We can add also something like: `Use the Yaml::DUMP_EXCEPTION_ON_INVALID_TYPE flag instead.`.

-> https://github.com/symfony/symfony/blob/v3.1.0/src/Symfony/Component/Yaml/Dumper.php#L65

greg0ire: Lol these people really like horizontal scroll :trollface: But, yeah, I'll add a sample recommendation.

greg0ire: Done

project/CONTRIBUTING.md
```diff
@@ -186,6 +186,25 @@ interface BarInterface
...
 ```
 
+In some cases, you will have the possibility to warn the user that things will change,
+and recommend a new way of doing things. You can do so by triggering the dedicated kind of error, like this:
+
+```php
+<?php
+if (/* some condition showing the user is using the legacy way */) {
+    @trigger_error(
+        'The '.__METHOD.' method is deprecated since 42.x, to be removed in 43.0. '.
```
Soullivaneuh: `__METHOD` -> `__METHOD__`

greg0ire: Woops. Fixed

Soullivaneuh:
Thank you @greg0ire. :+1:

greg0ire:
> Thank you @greg0ire. :+1:

My pleasure :)

Issue #116, opened 2016-05-30
Make abstract test case for user


soullivaneuh:
Lot of Sonata classes are abstract classes and the goal is to be extended on your projects.

I'm thinking about `AbstractAdmin` or `BlockService`.
## The test case

The goal of a test case is to provide facilities to the user to test project classes extending Sonata classes.

A test case can:
- Init needed properties on `setUp` and `setUpBeforeClass`
- Add some advanced `assert` methods
- Have some `test*` method that will be launched on each children test class.
## The name

I suggest to name those kind of class `Abstract<TargetClass>TestCase`.
- `Abstract` prefix because, well, it's an abstract class. :-)
- `TestCase` to differentiate internal abstract test class that should not be used by ending user and abstract class dedicated for this
## BC compliance

By definition, test classes should not be affected by the BC rule.

`*TestCase` classes must do, because they will be used.

Tag all other test classes with the `@internal` label can do the trick, but will be painful to apply.

So I'm thinking about the [`@api`](https://phpdoc.org/docs/latest/references/phpdoc/tags/api.html) tag.

> The @api tag represents those Structural Elements with a public visibility which are intended to be the public API components for a library or framework. Other Structural Elements with a public visibility serve to support the internal structure and are not recommended to be used by the consumer.
> 
> The exact meaning of Structural Elements tagged with @api MAY differ per project. It is however RECOMMENDED that all tagged Structural Elements SHOULD NOT change after publication unless the new version is tagged as breaking Backwards Compatibility.

This is exactly the goal here.

With that, we can quickly see which test class can be broke and which should not.

A documentation about that should be added somewhere for ending users.
## Sources

Class example from Symfony: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/Tests/HttpCache/HttpCacheTestCase.php

A class from one of my projects: https://github.com/Soullivaneuh/IsoCodesValidator/blob/3.x/tests/Constraints/AbstractConstraintValidatorTest.php

This is not a test case class, but you can see the principle of "default test" like `testNullIsValid`. This avoid test code duplication.

What is your opinion about this?

greg0ire:
I think there can be two sorts of abstract tests : 

1. internal tests
2. abstract tests the user can extend to tests their classes

I think both should be separated, because the first should not be autoloadable

[...]

PR #130, opened 2016-06-01
added .gitattributes file

### LINKED ISSUE #129: Do we need .gitattributes file?
Something like this:

```
.editorconfig    export-ignore
.gitattributes   export-ignore
.gitignore       export-ignore
CONTRIBUTING.md  export-ignore
```

Please use 👍  or 👎  to vote!

OskarStark:
Closes #129

### DIFF
--- a/project/.gitattributes.twig
+++ b/project/.gitattributes.twig
@@ -1,0 +1,3 @@
+.* export-ignore
+*.md export-ignore
+Tests/* export-ignore


project/.gitattributes
```diff
@@ -0,0 +1,4 @@
+.editorconfig    export-ignore
```
Soullivaneuh: Why not `.*`?

Soullivaneuh:
Could we add the documentation part?

project/.gitattributes
```diff
@@ -0,0 +1,6 @@
...
+.gitattributes   export-ignore
+.gitignore       export-ignore
+CONTRIBUTING.md  export-ignore
+
+/Resources/doc   export-ignore
```
Soullivaneuh: This is not always this path. Pleas use `{{ docs_path }}` instead.

Example: https://github.com/sonata-project/dev-kit/blob/ae18f3ea3755f543d83b4f713b752dc66398216d/project/Makefile#L17

greg0ire:
What about tests?

Soullivaneuh:
> What about tests?

I would like #116 consensus before work on this.

project/.gitattributes
OskarStark: we can...

project/.gitattributes
OskarStark: done

project/.gitattributes
OskarStark: done

OskarStark:
i got this error:

```
oskar.stark:/Volumes/development/workspaces/dev-kit (gitattributes-file)$ git rebase -i HEAD~2
}} is not a valid attribute name: project/.gitattributes:4
}} is not a valid attribute name: project/.gitattributes:4
}} is not a valid attribute name: project/.gitattributes:4
[losgelöster HEAD 4402d52] added .gitattributes file
 Date: Wed Jun 1 17:04:46 2016 +0200
 1 file changed, 4 insertions(+)
 create mode 100644 project/.gitattributes
Successfully rebased and updated refs/heads/gitattributes-file.
oskar.stark:/Volumes/development/workspaces/dev-kit (gitattributes-file)$
````

**Comment by greg0ire** [comment]
Maybe name it `.gitattributes.twig` ?

**Comment by OskarStark** [comment]
> Maybe name it .gitattributes.twig ? It will avoid git recognizing it for dev-kit

sound good, what do you think @Soullivaneuh ?

**Comment by Soullivaneuh** [comment]
@OskarStark Yes you can. In this case, the Twig variable name would be `docs_path`.

**Comment by OskarStark** [comment]
so `docs_path` instead of `{{ docs_path }}` ?

**Comment by Soullivaneuh** [comment]
@OskarStark Well, it's a Twig file! So AFAIK you still need to use `{{ }}`! :trollface:

**Comment by OskarStark** [comment]
> @OskarStark Well, it's a Twig file! So AFAIK you still need to use {{ }}! :trollface:

Exactly thats why i was asking! so now we should be fine! 👍

File: project/.gitattributes.twig [hunk 3]
```diff
@@ -0,0 +1,4 @@
+.*               export-ignore
+CONTRIBUTING.md  export-ignore
```
Soullivaneuh: `*.md`? Except if somebody want to keep a specific `.md` file...

OskarStark: > *.md? Except if somebody want to keep a specific .md file...

i like the idea 👍

OskarStark: done

OskarStark:
i removed the indention, because we don't know the length of the docs path....

Soullivaneuh:
> i removed the indention, because we don't know the length of the docs path....

Hmm... or we can just specify the two cases (`Resources/doc` and `docs`).

I don't think git will cry if a ignored path does not exist, will do?

OskarStark:
no it won't but the indention is anyway bad if we add something like `foo/bar/baz/loooonnnnggg` tomorrow, the the whole file will be modified in the diff, it is the exact same problem with the double arrow align

greg0ire:
What's wrong with the Twig variable?

Soullivaneuh:
> no it won't but the indention is anyway bad

Maybe, but I never seen a `.gitattributes` file without indentation. I don't know if a convention exists about that file...

OskarStark:
> Maybe, but I never seen a .gitattributes file without indentation. I don't know if a convention exists about that file...

https://git-scm.com/book/uz/v2/Customizing-Git-Git-Attributes#Merge-Strategies

greg0ire:
`man gitattributes`

OskarStark:
> What's wrong with the Twig variable?

Nothing i thought i should change `{{ docs_path }}` to `docs_path` without the curly braces... that confused me a few seconds 😄

greg0ire:
>     What's wrong with the Twig variable?

No I meant why not use it, but I get it know, it was all about alignment

OskarStark:
for example when we use this:

```
foo                      export-ignore
bar                      export-ignore
{{ docs_path}}     export-ignore
```

this would be the result:
```
foo                      export-ignore
bar                      export-ignore
/Resources/docs    export-ignore
```

you know what i mean?

greg0ire:
> you know what i mean?

Yes. Headaches ahead if we do that, as you explained.

Soullivaneuh:
@OskarStark Looks good. Can you try it on a one of your forked sonata project?

After that, I would like @rande opinion for this before merge because I got a refusal from him last time and he broke my heart. :broken_heart: :stuck_out_tongue:

greg0ire:
please

rande:
@Soullivaneuh You should also ask the community ;)

Soullivaneuh:
> @Soullivaneuh You should also ask the community ;)

Indeed, we might start a pool. I'll look for this week-end or next week. :+1:

Soullivaneuh:
> You should also ask the community ;)

After all, @OskarStark already start a vote.

@rande @sonata-project/contributors: https://twitter.com/soullivaneuh/status/738499577775566848

greg0ire:
Well the poll is overwhelmingly for this change, so RTM

Soullivaneuh:
@rande I let you merge this one.

Soullivaneuh:
@rande Ok to merge?

I would like to merge it as the same sime as the pending others.

I just need your approbation.

Soullivaneuh:
Found this: https://github.com/doctrine/collections/pull/79

Seems to be adopted by Doctrine too.

greg0ire:
@Soullivaneuh : It has been two weeks, I don't think @rande will hold a grudge against us if we merge this.

Soullivaneuh:
> @Soullivaneuh : It has been two weeks, I don't think @rande will hold a grudge against us if we merge this.

I'm too scary. :trollface:

rande:
I am against it if you exclude the Docs folder. Documentation is code too ...

Soullivaneuh:
> I am against it if you exclude the Docs folder. Documentation is code too ...

RST format is not the best way to read, online documentation is...

@OskarStark Can you remove doc from `.gitattributes` for now? We can merge the rest and open a new issue to vote about that.

OskarStark:
> @OskarStark Can you remove doc from .gitattributes for now? We can merge the rest and open a new issue to vote about that.

done @Soullivaneuh

Soullivaneuh:
Thanks @OskarStark, this will be merged soon.

If you want to, you could open an issue to discuss about documentation ignore with vote.

OskarStark:
i agree with @rande that documentation could be helpful in some rare cases, so for me this is fine as a first step!

core23:
Maybe we should remove `phpunit.xml.dist` and `Makefile` too

OskarStark:
sounds reasonable...

@greg0ire @Soullivaneuh what do you think?

project/.gitattributes.twig
```diff
@@ -0,0 +1,2 @@
+.* export-ignore
+*.md export-ignore
```
core23: When removing all `Upgrade*.md` and `Readme.md` files, we should also remove the docs.

greg0ire:
I agree!

Soullivaneuh:
> Maybe we should remove `phpunit.xml.dist` and `Makefile` too

Agree.

> When removing all `Upgrade*.md` and `Readme.md` files, we should also remove the docs.

Quite agree but let's discuss about this on another PR/issue! :+1:

OskarStark:
but, if we remove the `phpunit.xml.dist` shouldn't we remove the tests, too ❓

Soullivaneuh:
> but, if we remove the `phpunit.xml.dist` shouldn't we remove the tests, too ❓ 

We don't care on this case, it's applies only for tarball, not clones.

core23:
> but, if we remove the `phpunit.xml.dist` shouldn't we remove the tests, too ❓ 

We don't need the phpunit config outside the bundle, but some bundles depends on the `Tests/*.php` classes. Refs https://github.com/sonata-project/dev-kit/issues/179

Soullivaneuh:
> We don't need the phpunit config outside the bundle, but some bundles depends on the `Tests/*.php` classes. Refs https://github.com/sonata-project/dev-kit/issues/179

Yeah, but this should be removed, shouldn't be?

project/.gitattributes.twig
core23: >> When removing all `Upgrade*.md` and `Readme.md` files, we should also remove the docs.

> Quite agree but let's discuss about this on another PR/issue! :+1: 

Then we should remove this entry.

core23:
> Yeah, but this should be removed, shouldn't be?

Yes, but we can't remove them in all minor versions.

project/.gitattributes.twig
Soullivaneuh: @rande Talk about keeping the docs folder, not the md files. Plus, the `README.md` file is just a generic one with a link pointing to the online documentation.

So let's keep as is for now and launch a vote for docs folders.

Soullivaneuh:
> Yes, but we can't remove them in all minor versions.

Already discussed on the concerned issue: This have to be fixed internally on **minor** versions. We don't care about external usage because it should not be used at all.

core23:
We can now remove the `Tests` folder, it is excluded from the autoloader in all composer projects.

core23:
Ping @OskarStark

OskarStark:
done @core23

OskarStark:
I did everything, can you please merge this @Soullivaneuh ?

OskarStark:
Can we please merge this @Soullivaneuh ?

OskarStark:
ping @Soullivaneuh

core23:

core23:
Ping @Soullivaneuh

PR #154, opened 2016-06-09
Bump Symfony versions tests to 2.8

### LINKED ISSUE #51: Next major: Symfony version support
Let's talk about Symfony version support drop for next major of sonata bundles.

I think we all agree that SF 2.3 to 2.6 would be dropped.

What I'm not sure it's about 2.7. Should will still support it?

My thinking: Keeping only 2.8+ support would get us free of **LOT** of BC tricks between SF 2.7 and 2.8, especially for Form.

With that, we will have a clean code to start BC layers for next 3.x versions of Symfony.

I know, that 2.7 is a LTS version. But this one is supported by our actual bundles and SF 2.8 is also an LTS. BTW, 2.8 is here because of a too early release of 2.7 AFAIK.

Let's vote, use emoji reaction for that:

:+1: For support beginning at SF **2.8**
:-1: For support beginning at SF **2.7**

Feel free to elaborate your thinking on this issue. :+1: 

cc @sonata-project/contributors

Soullivaneuh:
Closes #51.

This will update Travis config **only**. Modifications have to be made on the `composer.json` of each project too.

@sonata-project/contributors If you have any suggestion/question, feel free! :+1:

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -3,7 +3,7 @@ admin-bundle:
     master:
       php: ['5.3', '5.4', '5.5', '5.6', '7.0']
       versions:
-        symfony: ['2.3', '2.7', '2.8', '3.0', '3.1']
+        symfony: ['2.8', '3.0', '3.1']
         sonata_core: ['3']
         sonata_block: ['3']
     3.x:
@@ -20,7 +20,7 @@ admin-search-bundle:
     master:
       php: ['5.3', '5.4', '5.5', '5.6', '7.0']
       versions:
-        symfony: ['2.3', '2.7', '2.8']
+        symfony: ['2.8']
         sonata_admin: ['3']
         ruflin_elastica: ['2']
 
@@ -29,7 +29,7 @@ block-bundle:
     master:
       php: ['5.3', '5.4', '5.5', '5.6', '7.0']
       versions:
-        symfony: ['2.3', '2.7', '2.8', '3.0', '3.1']
+        symfony: ['2.8', '3.0', '3.1']
         sonata_core: ['3']
         # https://travis-ci.org/sonata-project/SonataBlockBundle/jobs/131844587#L222
         #sonata_admin: ['3']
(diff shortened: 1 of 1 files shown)


core23:

OskarStark:
But we currently support Symfony 2.3, or not?

core23:
Yes, the stable still supports 2.3. This affects only the master branches.

core23:
Ping @Soullivaneuh ?

core23:

greg0ire:
If you want @Soullivaneuh to answer on a particular post, I think you should talk to him on slack.

greg0ire:
> This affects only the master branches.

Or does it?

config/projects.yml
```diff
@@ -259,7 +259,7 @@ media-bundle:
       php: ['5.3', '5.4', '5.5', '5.6', '7.0']
       target_php: 5.6
       versions:
-        symfony: ['2.3', '2.7', '2.8', '3.0', '3.1']
+        symfony: ['2.8', '3.0', '3.1']
```
greg0ire: @jordisala1991 : the mistake happened here. Can you do a PR on the admin bundle, just to see if the build would be green if we restored those?

config/projects.yml
```diff
@@ -259,7 +259,7 @@ media-bundle:
...
         doctrine_odm: ['1']
         sonata_core: ['3']
         sonata_admin: ['3']
```
greg0ire: Also, not sure what is the point of these lines…

jordisala1991: This line add a build on travis to check if we are compatible with 3.x

greg0ire: Yeah, so… what's the point? 3.x is the default I think, and is used on almost every other build

jordisala1991: when you build media-bundle 3.x you are using a tag of sonata-admin like 3.13.0. This allows to test against sonata-admin 3.x

greg0ire: Oh right…

PR #156, opened 2016-06-10
Add combine_consecutive_unsets fixer rule

Soullivaneuh:
Calling unset on multiple items should be done in one call.


Soullivaneuh:
Other fixers to add, will open another PR.

PR #164, opened 2016-06-11
Removed unstable branch for dashboard-bundle

core23:
As long as we don't have a stable release, we could use the master branch for development. 

We are <1.0, so we can't have any BC break, which would justify a separate branch.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -108,13 +108,6 @@ dashboard-bundle:
         sonata_core: ['3']
         sonata_admin: ['3']
         sonata_block: ['3']
-    0.x:
-      php: ['5.3', '5.4', '5.5', '5.6', '7.0']
-      versions:
-        symfony: ['2.3', '2.7', '2.8', '3.0', '3.1']
-        sonata_core: ['3']
-        sonata_admin: ['3']
-        sonata_block: ['3']
 
 datagrid-bundle:
   branches:


Soullivaneuh:
That change not really nothing. 1.x is the default branch, so let's work on this only and the backport will be automatically made on master.

BTW, doing this would produce weird results on documentation.

Soullivaneuh:
@core23 Can I close it or do you have anything to say about my comment?

core23:
I don't get the advantage of supporting a `0.x` / dev branch and a `master` branch for the moment. We should remove the `0.x` branch unless we have a stable 1.x release.

Soullivaneuh:
I get that at your first comment. What about my answer?

Sullivan SENECHAL
Le 15 juin 2016 19:24, "Christian Gripp" <notifications@github.com> a
écrit :

Soullivaneuh:
After reflection, this is the same case for admin-search-bundle.

I can do that, but all opened PR will be automatically closed.

Do you confirm @core23 ?

core23:
Maybe we should remove the branch on the admin search bundle first, because there are no open PRs.

For the dashboard bundle: Could we remove the master and rename the 0.x branch to master to port all PRs? Never tried :/

core23:
Or just wait until we accept all open PRs? What do you think @greg0ire

Soullivaneuh:
> Maybe we should remove the branch on the admin search bundle first, because there are no open PRs.

You didn't understand me, there is no 0.x branch on admin-search: https://github.com/sonata-project/SonataAdminSearchBundle/branches

Nothing to do here.

> For the dashboard bundle: Could we remove the master and rename the 0.x branch to master to port all PRs? Never tried :/

I don't think we can do that. "Renaming a branch" is the same as push a new one and delete the all.

@greg0ire Can you confirm?

greg0ire:
> I don't think we can do that. "Renaming a branch" is the same as push a new one and delete the all.

I think so too.

PR #151, opened 2016-06-09
Issue template rework

Soullivaneuh:

PR #171, opened 2016-06-15
use 'PHP version' instead of 'php version'

OskarStark:


Soullivaneuh:
The Issue template is reworked here: #151

OskarStark:
Sorry, i missed this!

PR #175, opened 2016-06-17
Getting a stack trace

greg0ire:
There are still a lot of issues without a stack trace, let's try to fix that.


Soullivaneuh:
Looks good.

PR #177, opened 2016-06-18
[WIP] Symfony way

soullivaneuh:
The goal is to take profit of Symfony services container and controllers.

I'll try to not introduce the full stack as we don't need it for this project.

cc @greg0ire 
- [ ] Use Symfony services on commands
- [ ] Introduce Symfony controllers
- [ ] Remove Silex
- [ ] Rework parameters and config (.env not needed) if necessary
- [ ] Add some unit and feature tests (behat)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
         "symfony/console": "^3.0",
         "symfony/filesystem": "^3.0",
         "twig/twig": "^1.24",
-        "vlucas/phpdotenv": "^2.2"
+        "vlucas/phpdotenv": "^2.2",
+        "symfony/http-kernel": "^3.1"
     },
     "require-dev": {
         "symfony/var-dumper": "^3.0"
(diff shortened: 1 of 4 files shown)


greg0ire:
"Freature" makes me hungry :trollface:

greg0ire:
How about a container interop implementation instead? Won't sf move to this one day?

Soullivaneuh:
> How about a container interop implementation instead? Won't sf move to this one day?

You got my attention, can you develop?

greg0ire:
I'm in the train right now so hard to develop but IIRC there is a PSR for containers, so let's use an implementation of this standard maybe?

Soullivaneuh:
You mean this one? https://github.com/php-fig/fig-standards/blob/master/proposed/container.md

Right, if you say Symfony will go for it, we can use it waiting that, no? Or have you other suggestions?

greg0ire:
Looks like it is not even accepted yet and symfony container already named its method the right way so let's use sf

core23:
Are you still working on this @Soullivaneuh ?

core23:
This PR looks dead. Feel free to reopen if you will continue working on this.

PR #178, opened 2016-06-18
Docker integration

Soullivaneuh:
To provide a "ready to use" dev env.

### DIFF
--- a/docker/conf/php.ini
+++ b/docker/conf/php.ini
@@ -1,0 +1,8 @@
+; PHP.ini required and recommended options.
+; Used by GitLab-CI and docker-compose
+
+[PHP]
+memory_limit = -1
+
+[Date]
+date.timezone = Europe/Paris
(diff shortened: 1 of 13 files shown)


docker/conf/php.ini
```diff
@@ -0,0 +1,8 @@
+; PHP.ini required and recommended options.
+; Used by GitLab-CI and docker-compose
+
+[PHP]
+memory_limit = -1
+
+[Date]
+date.timezone = Europe/Paris
```
greg0ire: Not sure about this :P

Soullivaneuh: Well, I set this to not get PHP yelling.

But this can indeed be great if it can be configurable.

Any suggestion?

greg0ire:
Utc?

Soullivaneuh:
> Utc?

Well, depends of where you are, right? :stuck_out_tongue:

greg0ire:
Utc means Coordinated **Universal** Time, so no, it does not depend, it applies to the whole Earth (and even the universe ?).

PR #191, opened 2016-08-02
Fix reversed .env file names on configure

Enleur:
Error on configure step


Soullivaneuh:
Nice catch, thank you!

PR #193, opened 2016-08-12
classification-media-bundle

core23:
Setup for the new classification media bridge.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -80,6 +80,13 @@ classification-bundle:
         symfony: ['2.3', '2.7', '2.8', '3.0', '3.1']
         sonata_admin: ['3']
 
+classification-media-bundle:
+  branches:
+    master:
+      php: ['5.3', '5.4', '5.5', '5.6', '7.0']
+      versions:
+        symfony: ['2.3', '2.7', '2.8', '3.0', '3.1']
+
 comment-bundle:
   branches:
     master:


config/projects.yml
```diff
@@ -80,6 +80,13 @@ classification-bundle:
         symfony: ['2.3', '2.7', '2.8', '3.0', '3.1']
         sonata_admin: ['3']
 
+classification-media-bundle:
+  branches:
+    master:
+      php: ['5.6', '7.0']
```
Soullivaneuh: Didn't take a look to the dependence separation, but it would be great to have a BC way.

In this case, we should support older dependencies, shouldn't we?

core23: No, because the new bundle would only support the latest master/4.0 version of the dependencies.

Supporting the 3.x branches too would be a lot of pain...

core23:
Ping @Soullivaneuh

config/projects.yml
Soullivaneuh: > No, because the new bundle would only support the latest master/4.0 version of the dependencies.

This is especially what I'm talking about. :-)

Soullivaneuh: > No, because the new bundle would only support the latest master/4.0 version of the dependencies.

This is especially what I'm talking about. :-)

Moving code to a separated bundle is BC break, right? I think this can be BC with ease:

* Move the code (already done)
* Deprecate the copied classes on 3.x and require the new bundle (BC part)
* Then on 4.x, remove the classes

core23:
Pong @Soullivaneuh

Soullivaneuh:
@core23 I need at least a `composer.json` file on the repo: https://github.com/sonata-project/SonataClassificationMediaBundle

Then we can add it on Packagist to make dev-kit working.

Sorry for the delay.

core23:
Done 👍

core23:
Pung @Soullivaneuh

core23:
Any news on the setup @Soullivaneuh ?

core23:
@rande Can you please add the new repo to packagist? After that we can merge this PR

core23:
Packagist is setup, so RTM!

greg0ire:
Let's try to make it green first, shall we?

config/projects.yml
```diff
@@ -80,6 +80,13 @@ classification-bundle:
         symfony: ['2.3', '2.7', '2.8', '3.0', '3.1']
         sonata_admin: ['3']
 
+classification-media-bundle:
+  branches:
+    master:
+      php: ['5.3', '5.4', '5.5', '5.6', '7.0']
```
greg0ire: Why so low?

config/projects.yml
```diff
@@ -80,6 +80,13 @@ classification-bundle:
...
+      versions:
+        symfony: ['2.3', '2.7', '2.8', '3.0', '3.1']
```
greg0ire: Same here, this is master.

config/projects.yml
core23: Because we do use this on the stable branches :(

greg0ire: Oh you mean code here will be coming from media / classification, and a BC solution should be provided? But this is `master`, right? So you can drop them anyway, at any time, so why not now? Also, why is there only a definition for the master branch?

core23: > Oh you mean code here will be coming from media / classification, and a BC solution should be provided?

That's the idea. I'll try to make a BC version for the stable classification bundle, so we need the lower deps for PHP. 

Maybe we could create a 0.x release for the new bundle for a short time. So we can drop it later with the 1.0 release.

> Also, why is there only a definition for the master branch?

Because it's the initial setup. We should we create two branches for this?

greg0ire: Because you are going to try to be BC with something on one branch, but what if someone wants to contribute a BC-breaking change?

core23: After releasing a stable version, we could add the branch system

PR #207, opened 2016-09-24
Document when and how to use @deprecated

### LINKED ISSUE #206: deprecated vs NEXT_MAJOR
Classes and method might be marked as deprecated. @core23 suggested that in that case, `NEXT_MAJOR` would be somewhat redundant. What should we do ? 

Things to be taken into account : 
- Always using `NEXT_MAJOR` will make greping for things left to do easier
- [@core23 couldn't find any usage of both things together](https://github.com/sonata-project/SonataCoreBundle/pull/311#r79307431)
- `@deprecated` plays nice with IDEs

greg0ire:
Closes #206

### DIFF
--- a/project/CONTRIBUTING.md
+++ b/project/CONTRIBUTING.md
@@ -208,6 +208,21 @@ if (/* some condition showing the user is using the legacy way */) {
 }
 ```
 
+Additionally, and when applicable, you must use the `@deprecated` tag on classes or methods you wish to deprecate,
+along with a message directed at the end user (as opposed to other contributors).
+
+
+```php
+/**
+ * NEXT_MAJOR: remove this method
+ *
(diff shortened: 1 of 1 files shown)


project/CONTRIBUTING.md
```diff
@@ -208,6 +208,21 @@ if (/* some condition showing the user is using the legacy way */) {
...
 
+Additionally, and when applicable, you must use the `@deprecated` tag on classes or methods you wish to deprecate,
+along with a message directed at the end user (as opposed to other contributors).
+
+
+```php
+/**
+ * NEXT_MAJOR: remove this method
+ *
+ * @deprecated since 3.x, to be removed in 4.0. Use Foo::bar instead
```
Soullivaneuh: Missing a dot at the end of the sentence.

greg0ire: Sorry, fixed.

PR #213, opened 2016-11-12
Added assets task to Makefile

core23:
Added a new task to build assets. These task are used in:

https://github.com/sonata-project/SonataPageBundle/tree/3.x/Resources/assets_src
https://github.com/sonata-project/SonataDashboardBundle/tree/master/Resources/assets_src


project/Makefile
```diff
@@ -20,3 +20,7 @@ test:
 
 docs:
 	cd {{ docs_path }} && sphinx-build -W -b html -d _build/doctrees . _build/html
+
+assets:
+	cd Resources/assets_src && npm install
+	cd Resources/assets_src && gulp
```
greg0ire: So… `cd Resources/assets_src` twice? This does not look right. Maybe do this instead : 

```bashrc
pushd Resources/assets_src && npm install && popd
pushd Resources/assets_src && gulp && popd
```

core23:
Added a new config key for the `assets_path`

core23:
@sonata-project/contributors Can you please give another review and merge?

OskarStark:
> @sonata-project/contributors Can you please give another review and merge?

sorry I am not really firm with this topic

project/Makefile
```diff
@@ -20,3 +20,8 @@ test:
 
 docs:
 	cd {{ docs_path }} && sphinx-build -W -b html -d _build/doctrees . _build/html
+
+assets:
+	pushd {{ assets_path }} && npm install && popd
```
Soullivaneuh: Is this concerning all projects? Because it will be deployed on each of them.

core23: ATM only the dashboard and page bundle...

OskarStark: in this case we should to this only for these bundles IMO

core23: How can I do this?

OskarStark: IMO we need to add this to this file, but interpret a new config value/variable if this should be added to the used bundle

@Soullivaneuh what do you think and how should we achive this or is it not necessary?

project/Makefile
```diff
@@ -20,3 +20,9 @@ test:
 
 docs:
 	cd {{ docs_path }} && sphinx-build -W -b html -d _build/doctrees . _build/html
+
+{% if assets_path is not empty %}
```
Soullivaneuh: How is the rendered Makefile? Maybe we have to deal with withespaces here.

Soullivaneuh: `if assets_path` would be enough IMHO.

core23:
Fixed that

project/Makefile
OskarStark: what about:

```twig
{-% if assets_path is not empty %-}

{-% endif %-}
```

core23: How can I manually run the dev-kit build?

Soullivaneuh:
@core23 Still not see the related config files... :wink:

core23:
Not sure what config files you are looking for

Soullivaneuh:
@core23 You configured an `assets_path` config key but this one is not used at all.

No modification was made on the [`projects.yml`](https://github.com/sonata-project/dev-kit/blob/c5b400abff42957e8252c4390fc8423887df65cc/config/projects.yml) file.

core23:
You're right, the most important part was missing xD

core23:

core23:
Anything to do here?

core23:
Ping @Soullivaneuh

project/Makefile.twig
```diff
@@ -20,3 +20,9 @@ test:
 
 docs:
 	cd {{ docs_path }} && sphinx-build -W -b html -d _build/doctrees . _build/html
+{% if assets_path %}
+
+assets:
+	pushd {{ assets_path }} && npm install && popd
+	pushd {{ assets_path }} && gulp && popd
```
greg0ire: missing `.PHONY: assets`

core23: Fixed

project/Makefile.twig
```diff
@@ -20,3 +20,9 @@ test:
 
 docs:
 	cd {{ docs_path }} && sphinx-build -W -b html -d _build/doctrees . _build/html
+{% if assets_path %}
+
+assets:
```
greg0ire: Would be great to do something like this, or even move to yarn and do exactly this: https://stackoverflow.com/questions/44036997/how-to-prevent-yarn-install-from-running-twice-in-makefile#44226605

src/Console/Command/DispatchCommand.php
```diff
@@ -214,6 +214,7 @@ private function updateDevKitHook(Package $package)
             if (array_key_exists('url', $hook['config'])
                 && 0 === strncmp($hook['config']['url'], $hookBaseUrl, strlen($hookBaseUrl))) {
                 $devKitHook = $hook;
+
```
Soullivaneuh: This should not be here.

project/Makefile.twig
```diff
@@ -2,7 +2,7 @@
 #
 # It's auto-generated by sonata-project/dev-kit package.
 
-.PHONY: test docs
+.PHONY: test docs{% if assets_path %} assets{% endif %}
```
Soullivaneuh: It would be great to avoid the twig template of the Makefile.

Could we play another way? Maybe generic setup task and/or env vars?

project/Makefile.twig
core23: Yarn is not a build system. It's a tool for dependency management.

greg0ire: Now that I think of it, having a `.lock` in a library is probably not a good idea anyway

project/Makefile.twig
core23: Maybe switch to [composer scripts](https://github.com/sonata-project/dev-kit/issues/263)?

project/Makefile.twig
greg0ire: But I don't know how npm and yarn differ so much that it's ok to use one and not the other.

core23: Changed `npm` to `yarn`, because this is a lot faster. A `.lock` file is optional (same as composer.lock). So this is okay

project/Makefile.twig
Soullivaneuh: This is front building.

Why not make a yarn custom script and call it on travis bash files?

core23:
Can we please merge this @Soullivaneuh ?

This might not be the best solution, but it is a quick for for building the assets.

Soullivaneuh:
What about my comment? https://github.com/sonata-project/dev-kit/pull/213/files#r140774356

BTW, the makefile already use some var replacing, I'm not sure it will work properly with the twig mode.

project/Makefile.twig
core23: Will this change anything? We still need to call `yarn run assets`.

The `Makefile` is our main file for all automated jobs (e.g. phpunit, docs, etc.), so we should add this yarn stuff too.

Soullivaneuh: > Will this change anything?

Yes: https://github.com/sonata-project/dev-kit/pull/213/files#r140481910

PR #224, opened 2017-01-06
Show only outdated

greg0ire:
To be reviewed commit by commit ;)

The goal of this PR is to make issues easier to report, but also easier to read.

# TODO

- [x] remove third commit
- [x] replace `--outdated` with `--all` option (but keep `outdated` command)

### DIFF
--- a/project/.github/ISSUE_TEMPLATE.md
+++ b/project/.github/ISSUE_TEMPLATE.md
@@ -12,19 +12,19 @@
 
 #### Sonata packages
 
-```bash
+```
 $ composer show --latest 'sonata-project/*'
 ```
 
 #### Symfony packages
 
-```bash
+```
 $ composer show --latest 'symfony/*'
 ```
 
 #### PHP version
 
-```bash
+```
 $ php -v
 # Put the result here.
 ```
--- a/project/.github/ISSUE_TEMPLATE.md
+++ b/project/.github/ISSUE_TEMPLATE.md
@@ -13,15 +13,13 @@
 #### Sonata packages
 
 ```bash
-$ composer show sonata-project/*
-# Put the result here.
+$ composer show --latest sonata-project/*
 ```
 
 #### Symfony packages
 
 ```bash
-$ composer show symfony/*
-# Put the result here.
+$ composer show --latest symfony/*
 ```
 
 #### PHP version
--- a/project/.github/ISSUE_TEMPLATE.md
+++ b/project/.github/ISSUE_TEMPLATE.md
@@ -13,13 +13,13 @@
 #### Sonata packages
 
 ```bash
-$ composer show --latest sonata-project/*
+$ composer show --latest 'sonata-project/*'
 ```
 
 #### Symfony packages
 
 ```bash
-$ composer show --latest symfony/*
+$ composer show --latest 'symfony/*'
 ```
 
 #### PHP version


Soullivaneuh:
I'm not sure of this.

Listing sonata packages also permits to know which installation the user are, and not only updated.

Plus, when we will release the next major, the old one will be maintained for bug fixes.

So we have to know which versions is used, outdated or not.

greg0ire:
You're right. I can replace the `--outdated` option, but then only users will benefit from the highlight, because you obviously can't copy colors to github. But that still would be something I guess.

core23:
👎 Please don't forget the windows users! Not sure if the new docs will fit there.

greg0ire:
@core23 oh my god you're right, I totally forgot about them. But the script would still work, wouldn't it?

project/.github/ISSUE_TEMPLATE.md
```diff
@@ -10,25 +10,26 @@
...
-# Put the result here.
-```
-
-#### PHP version
-
-```bash
-$ php -v
-# Put the result here.
+outdatedSonata=$(composer outdated --outdated 'sonata-project/*')
+if [ "$outdatedSonata" != "" ]
```
core23: Windows should have problems with this stuff

greg0ire: I'll try to find a portable syntax, but don't hesitate if you know one already

project/.github/ISSUE_TEMPLATE.md
```diff
@@ -10,25 +10,26 @@
...
+then
+	echo 'Sonata Packages'
+	echo '---------------'
+	printf "%s\n\n" "$outdatedSonata"
```
Soullivaneuh: Same here for windows. It's `\r\n` AFAIK.

core23:
IMHO we should keep the old docs, because they are simple and everybody understands what `composer show` does.

greg0ire:
> IMHO we should keep the old docs, because they are simple and everybody understands what composer show does.

And I think it's great to learn new things, and that some of our users really need to learn about `composer outdated`

core23:
I've no problem with the `oudated` option, but the user should just run a  script which looks simple at first sight.

greg0ire:
Ok, I think I will just remove the fourth commit then.

greg0ire:
Added a TODO list

greg0ire:
Turns out `--outdated` is the default, you need `--all` to see all packages.

project/.github/ISSUE_TEMPLATE.md
```diff
@@ -12,21 +12,20 @@
 
 #### Sonata packages
 
-```bash
-$ composer show sonata-project/*
-# Put the result here.
+```
+$ composer outdated --all 'sonata-project/*'
```
Soullivaneuh: According to the outdated help description:

> The outdated command is just a proxy for `composer show -l`

So we can simply do:

```bash
composer show --latest 'sonata-project/*'
```

project/.github/ISSUE_TEMPLATE.md
```diff
@@ -12,21 +12,20 @@
...
+# Put the result here if there is one, otherwise, just remove this paragraph.
 ```
 
 #### Symfony packages
 
-```bash
-$ composer show symfony/*
-# Put the result here.
+```
+$ composer outdated --all 'symfony/*'
```
Soullivaneuh:
```bash
composer show --latest 'symfony/*'
```

project/.github/ISSUE_TEMPLATE.md
```diff
@@ -12,21 +12,20 @@
 
 #### Sonata packages
 
-```bash
-$ composer show sonata-project/*
-# Put the result here.
+```
+$ composer outdated --all 'sonata-project/*'
+# Put the result here if there is one, otherwise, just remove this paragraph.
```
Soullivaneuh: Why if there is one? It should always be AFAIK! :wink:

project/.github/ISSUE_TEMPLATE.md
```diff
@@ -12,21 +12,20 @@
...
-```bash
-$ composer show symfony/*
-# Put the result here.
+```
+$ composer outdated --all 'symfony/*'
 ```
 
 #### PHP version
 
-```bash
```
Soullivaneuh: This should not be remove. This is indeed bash syntax.

project/.github/ISSUE_TEMPLATE.md
greg0ire: Yup. Forgot to remove this.

project/.github/ISSUE_TEMPLATE.md
greg0ire: No it should really be removed. Don't forget, in the end, we will get the output of the composer, which is certainly not a bash script inside this fenced block.

Soullivaneuh: >  Don't forget, in the end, we will get the output of the composer, which is certainly not a bash script inside this fenced block.

First:

```bash
$ ./docker-console composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              3.10.3 3.10.3 The missing Symfony Admin Generator
sonata-project/block-bundle              3.2.0  3.2.0  Symfony SonataBlockBundle
sonata-project/cache                     1.0.7  1.0.7  Cache library
sonata-project/core-bundle               3.1.2  3.1.2  Symfony SonataCoreBundle
sonata-project/doctrine-extensions       1.0.2  1.0.2  Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.1.2  3.1.2  Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/easy-extends-bundle       2.1.10 2.1.10 Symfony SonataEasyExtendsBundle
sonata-project/exporter                  1.7.0  1.7.0  Lightweight Exporter library
sonata-project/intl-bundle               2.2.4  2.2.4  Symfony SonataIntlBundle
```

This works great, and this is a bash command, I really don't see the issue.

Second, you're talking about composer, but you modified the `php -v` command. :stuck_out_tongue:

greg0ire: Works the same, php -v doesn't output bash.

Soullivaneuh: Ok since your sample: https://github.com/sonata-project/SonataAdminBundle/issues/4267

Can you please do the same for composer outputs in this case?

greg0ire:
@Soullivaneuh please approve and merge. Also : https://github.com/composer/composer/commit/d6ffe1fd1df1c3d4052354af008d67d322399efd 😄

PR #226, opened 2017-01-17
show only on line if stable and unstable branch are equal

OskarStark:
We show the same branch twice in SonataAdminSearchBundle:

bildschirmfoto 2017-01-17 um 11 06 03

### DIFF
--- a/project/README.md
+++ b/project/README.md
@@ -13,7 +13,9 @@
 Branch | Travis | Coveralls |
 ------ | ------ | --------- |
 {{ stable_branch }}   | [![Build Status][travis_stable_badge]][travis_stable_link]     | [![Coverage Status][coveralls_stable_badge]][coveralls_stable_link]     |
-{{ unstable_branch }} | [![Build Status][travis_unstable_badge]][travis_unstable_link] | [![Coverage Status][coveralls_unstable_badge]][coveralls_unstable_link] |
+{% if unstable_branch != stable_branch %}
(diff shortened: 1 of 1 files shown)


project/README.md
```diff
@@ -13,7 +13,9 @@
 Branch | Travis | Coveralls |
 ------ | ------ | --------- |
 {{ stable_branch }}   | [![Build Status][travis_stable_badge]][travis_stable_link]     | [![Coverage Status][coveralls_stable_badge]][coveralls_stable_link]     |
-{{ unstable_branch }} | [![Build Status][travis_unstable_badge]][travis_unstable_link] | [![Coverage Status][coveralls_unstable_badge]][coveralls_unstable_link] |
+{% if unstable_branch != stable_branch %}
```
Soullivaneuh: This is very great except... this is not a Twig file! :trollface: 

And it would be great to keep it as is to keep Markdown preview with ease.

OskarStark: Oh I thought we were using twig to generate the README.md

What shall I change? I don't get it

OskarStark: otherwise we could release 0.1 of admin search bundle

greg0ire: Isn't 1.0 released?

OskarStark: Ah you are right....
so only Article and dashboard bundle are affected, but DashboardBundle already has a 0.1.0 release .. 😕  ❓

Soullivaneuh: I think this is not a very big deal to have duplicate information.

It will be cumbersome to introduce twig on mardown file and lose github preview just for that.

OskarStark: Ok, fine!

PR #234, opened 2017-02-07
sonata phpcr-odm admin 2 still supports php 5.6

dbu:
according to the composer.json, sonata phpcr-odm still supports php 5 versions. cmf 2.0 is going for php 5.6. i would like to do the same with the sonata admin, if that makes sense.

we rely on admin-bundle ^3.1 - it looks like that allows anything from php 5.3 and 7, so should work out, right?

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -174,7 +174,8 @@ doctrine-orm-admin-bundle:
 doctrine-phpcr-admin-bundle:
   branches:
     master:
-      php: ['7.0', '7.1']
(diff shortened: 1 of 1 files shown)


dbu:
thanks, added a comment to explain why we still want 5.6. i do not have merge rights here. if you can agree with this, please merge.

greg0ire:
I agree, but this is the only repo where I don't have merge rights. See with @Soullivaneuh .

Soullivaneuh:
@dbu Merged and deployed.

This will be applied this night.

dbu:
thanks!

PR #192, opened 2016-08-02
Add mongodb adapter to mongo build

enleur:
Use mongodb adapter in related builds

Related to https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/183
Fixes https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/186

PS I didn't check that it is working properly. Could you?

PR #238, opened 2017-02-25
Add mongodb to PHP7 with ext-mongo compatibility

jordisala1991:
Add mongodb extension to PHP7. With compatibility with old ext-mongo.

There are two issues that need to run a previous composer update to create composer.lock in order to not fail installing with correct composer flags.

https://github.com/composer/composer/issues/5355 that shows when using `prefer-lowest`
https://github.com/composer/composer/issues/5030 when using packages that provides extension implementations like: alcaeus/mongo-php-adapter

This is needed to start moving to PHP7.0 in our builds and can also be used to stop testing on PHP5.6 on 3.x branches.

There is a little overhead when running always two times composer update, but is not that big. Only if you are running PHP < 7, but still not that big difference.

This PR shows an example on media-bundle: https://github.com/sonata-project/SonataMediaBundle/pull/1211

Closes https://github.com/sonata-project/dev-kit/pull/192

### DIFF
--- a/project/.travis/before_install_test.sh.twig
+++ b/project/.travis/before_install_test.sh.twig
@@ -6,9 +6,16 @@ if [ "${TRAVIS_PHP_VERSION}" != "hhvm" ]; then
     TRAVIS_INI_FILE="$PHP_INI_DIR/travis.ini"
     echo "memory_limit=3072M" >> "$TRAVIS_INI_FILE"
 
+    {% if 'mongodb' in services %}
     if [ "$TRAVIS_PHP_VERSION" '<' '7.0' ]; then
         echo "extension=mongo.so" >> "$TRAVIS_INI_FILE"
+    else
+        echo "extension=mongodb.so" >> "$TRAVIS_INI_FILE"
+
+        # Backwards compatibility with old mongo extension
+        composer require "alcaeus/mongo-php-adapter" --no-update
     fi
+    {% endif %}
 fi
 
 # To be removed when following PR will be merged: https://github.com/travis-ci/travis-build/pull/718
--- a/project/.travis/install_test.sh
+++ b/project/.travis/install_test.sh
@@ -16,8 +16,9 @@ chmod u+x "${HOME}/bin/phpunit"
 wget https://github.com/satooshi/php-coveralls/releases/download/v1.0.1/coveralls.phar --output-document="${HOME}/bin/coveralls"
 chmod u+x "${HOME}/bin/coveralls"
 
-# To be removed when this issue will be resolved: https://github.com/composer/composer/issues/5355
-if [ "${COMPOSER_FLAGS}" = '--prefer-lowest' ]; then
-    composer update --prefer-dist --no-interaction --prefer-stable --quiet
-fi
+# To be removed when these issues are resolved:
+# https://github.com/composer/composer/issues/5355
+# https://github.com/composer/composer/issues/5030
+composer update --prefer-dist --no-interaction --prefer-stable --quiet --ignore-platform-reqs
+
 travis_wait composer update --prefer-dist --no-interaction --prefer-stable ${COMPOSER_FLAGS}


greg0ire:
FYI, I hunted for a place to store comments in composer.json, looks like [there is one now](https://github.com/composer/composer/blob/6c51006309c1692351f4e7f8395616a5946f5e42/CHANGELOG.md) : `_comment`

project/.travis/install_test.sh
```diff
@@ -16,8 +16,9 @@ chmod u+x "${HOME}/bin/phpunit"
 wget https://github.com/satooshi/php-coveralls/releases/download/v1.0.1/coveralls.phar --output-document="${HOME}/bin/coveralls"
 chmod u+x "${HOME}/bin/coveralls"
 
-# To be removed when this issue will be resolved: https://github.com/composer/composer/issues/5355
-if [ "${COMPOSER_FLAGS}" = '--prefer-lowest' ]; then
-    composer update --prefer-dist --no-interaction --prefer-stable --quiet
-fi
+# To be removed when this issues will be resolved:
```
greg0ire: "when these issues are resolved"

project/.travis/install_test.sh
```diff
@@ -16,8 +16,9 @@ chmod u+x "${HOME}/bin/phpunit"
...
+# https://github.com/composer/composer/issues/5355
+# https://github.com/composer/composer/issues/5030
+composer update --prefer-dist --no-interaction --prefer-stable --ignore-platform-reqs --quiet
```
greg0ire: Why quiet? It would make troubleshooting problems a bit hard…

project/.travis/install_test.sh
jordisala1991: Changed

project/.travis/install_test.sh
jordisala1991: Yes, it should not be a problem if we remove --quiet.

project/.travis/install_test.sh
```diff
@@ -16,8 +16,9 @@ chmod u+x "${HOME}/bin/phpunit"
 wget https://github.com/satooshi/php-coveralls/releases/download/v1.0.1/coveralls.phar --output-document="${HOME}/bin/coveralls"
 chmod u+x "${HOME}/bin/coveralls"
 
-# To be removed when this issue will be resolved: https://github.com/composer/composer/issues/5355
-if [ "${COMPOSER_FLAGS}" = '--prefer-lowest' ]; then
-    composer update --prefer-dist --no-interaction --prefer-stable --quiet
-fi
+# To be removed when these issues are resolved:
```
core23: Maybe add a `TODO` here.

greg0ire:
@jordisala1991 can you fix the conflicts?

project/.travis/before_install_test.sh.twig
```diff
@@ -5,9 +5,16 @@ if [ "${TRAVIS_PHP_VERSION}" != "hhvm" ]; then
...
     echo "memory_limit=3072M" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 
+    {% if 'mongodb' in services %}
     if [ ${TRAVIS_PHP_VERSION} '<' '7.0' ]; then
         echo "extension=mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+    else
+        echo "extension=mongodb.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+
+        # Backwards compatibility with old mongo extension
+        composer require "alcaeus/mongo-php-adapter" --no-update
```
Soullivaneuh: Why requiring this package here? Should not be added directly on the `composer.json` project file?

greg0ire: @Soullivaneuh this lib is only required if you use php7. Not sure what happens if you don't but it makes sense. Can you please resolve the conflicts, merge and deploy? I'm afraid @jordisala1991 is unavailable ATM and as you know, this is a blocker the next major release of exporter.

jordisala1991: It provides compatiblity with old PHP, because the extension name has changed. You can't add directly on composer.json require or require-dev because it depends on php version. Usually it is on suggest section, but that is not enough

project/.travis/install_test.sh
jordisala1991: Was trying to keep the same style as other comments on those files. I can change on all, but not sure if I should on this PR.

jordisala1991:
Rebased and fixed conflicts :)

project/.travis/before_install_test.sh.twig
```diff
@@ -6,9 +6,16 @@ if [ "${TRAVIS_PHP_VERSION}" != "hhvm" ]; then
...
     echo "memory_limit=3072M" >> "$TRAVIS_INI_FILE"
 
+    {% if 'mongodb' in services %}
     if [ "$TRAVIS_PHP_VERSION" '<' '7.0' ]; then
         echo "extension=mongo.so" >> "$TRAVIS_INI_FILE"
+    else
+        echo "extension=mongodb.so" >> "$TRAVIS_INI_FILE"
+
+        # Backwards compatibility with old mongo extension
+        composer require "alcaeus/mongo-php-adapter" --no-update
```
Soullivaneuh: Still thinking this can be added to `require-dev` of the project. Why not? This is also compatible with PHP 5.x.

jordisala1991: I just tried it on SonataMediaBundle (affected by this PR), and here are the results:

https://travis-ci.org/sonata-project/SonataMediaBundle/builds/250476334?utm_source=github_status&utm_medium=notification

You can't install it on composer.json because it requires atleast PHP 5.6 and we are building with PHP 5.3, 5.4 and 5.5. There is no version of `alcaeus/mongo-php-adapter` compatible with PHP < 5.5.

If we don't want to require it like in this PR the only way is to assume we can't do it for 3.x branch and only do it for master, but IMO the are some reasons to be able to do it like this:

* Faster builds (PHP7 vs PHP5)
* SonataMongoDBAdminBundle is not building with PHP7 even if it has the requirement on composer.json
* We can drop this code once the old extension is not used anymore on any of our projects
* It does not hurt a lot because it is only being required if mongodb is defined as a service.

PR #240, opened 2017-02-27
Remove composer self update from travis build process

jordisala1991:
This PR (https://github.com/travis-ci/travis-build/pull/718) is not merged yet, but the issue is solved now, you can see it here:

https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/php.rb#L165

And on every travis build, there is a triple composer update actually, two of them comes from travis itself and the last one is from us, those lines are no longer needed.

### DIFF
--- a/project/.travis/before_install_test.sh.twig
+++ b/project/.travis/before_install_test.sh.twig
@@ -10,8 +10,6 @@ if [ "${TRAVIS_PHP_VERSION}" != "hhvm" ]; then
     fi
 fi
 
-# To be removed when following PR will be merged: https://github.com/travis-ci/travis-build/pull/718
-composer self-update --stable
 sed --in-place "s/\"dev-master\":/\"dev-${TRAVIS_COMMIT}\":/" composer.json
 
 {% for package_name,package_versions in versions if package_versions|length > 0 %}


Soullivaneuh:
@jordisala1991 Seems your code link is not good anymore. Could you please provide a fixed one?

jordisala1991:
https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/php.rb#L178 This one @Soullivaneuh ?

Soullivaneuh:
Commit locked link are safer @jordisala1991. ;-)

https://github.com/travis-ci/travis-build/blob/9c637df2d205c5012ea88a6eee0a77f57672b03e/lib/travis/build/script/php.rb#L178

And yes the composer binary is updated, but not to a stable version. This is the goal of my not merged PR.

jordisala1991:
I dont understand. Look at a random travis build:

https://travis-ci.org/sonata-project/SonataMediaBundle/jobs/205571601#L152

This line updates already composer to latest version. We are doing another composer self-update that does nothing.

jordisala1991:
From composer (https://getcomposer.org/download/):

> By default the installer and composer self-update will download stable versions only. If you would like to help test pre-release versions you can use the --preview flag on either installer or self-update. For snapshot builds which are done from the latest Composer commit, you can use the --snapshot flag.

Soullivaneuh:
Ok this line works now: https://github.com/travis-ci/travis-build/blob/e5a45cbf49e0d9e27398e76e5f25dd7706feb6aa/lib/travis/build/script/php.rb#L182

PR #243, opened 2017-03-17
Release command 🚀

jordisala1991:
Command to make releases easy.

## ToDo

- [x] Compile changelog
- [x] List PR
- [x] get labels of all PR
- [x] Determine what is the next tag
- [x] Check if the base branch is okay to do a release: build is okay
- [x] Print labels with colors, mark the ones that are not supposed to be there
- [x] Check if changelog is present on all PR that need to have it
- [x] Cleanup code

## Nice to have
- [ ] Move some responsabilities out of the command
- [ ] Add test

## Subject

There are a lot of manual tasks that can be done automatically. Maybe we don't start with a full automatic release process, but this will be the first step.

- [x] Check if we can create a commit with the changelog modifications and create a tag and a release on the repository

Yes, we can:

https://developer.github.com/v3/repos/contents/#get-contents
https://developer.github.com/v3/repos/contents/#update-a-file
https://developer.github.com/v3/git/tags/#create-a-tag-object
https://developer.github.com/v3/repos/releases/#create-a-release

### DIFF
--- a/src/Command/ReleaseCommand.php
+++ b/src/Command/ReleaseCommand.php
@@ -1,0 +1,307 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Command;
+
+use Packagist\Api\Result\Package;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+
+/**
+ * @author Jordi Sala <jordism91@gmail.com>
+ */
+final class ReleaseCommand extends AbstractCommand
+{
+    private static $labels = [
+        'patch' => 'blue',
+        'bug' => 'red',
+        'docs' => 'yellow',
+        'minor' => 'green',
+        'pedantic' => 'cyan',
+    ];
+
+    private static $stabilities = [
+        'patch' => 'blue',
+        'minor' => 'green',
+        'pedantic' => 'yellow',
+    ];
+
+    protected function configure(): void
+    {
+        parent::configure();
+
+        $help = <<<'EOT'
+The <info>release</info> command analyzes pull request of a given project to determine
+the changelog and the next version to release.
+
+Usage:
+
+<info>php dev-kit release</info>
+
+First, a question about what bundle to release will be shown, this will be autocompleted will
+the projects configured on <info>projects.yml</info>
+
+The command will show what is the status of the project, then a list of pull requests
+made against selected branch (default: stable branch) with the following information:
+
+stability, name, labels, changelog, url.
+
+After that, it will show what is the next version to release and the changelog for that release.
+EOT;
+
+        $this
+            ->setName('release')
+            ->setDescription('Helps with a project release.')
+            ->setHelp($help);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $project = $this->getProject($input, $output);
+        $branches = array_keys($this->configs['projects'][$project]['branches']);
+        $branch = \count($branches) > 1 ? next($branches) : current($branches);
+
+        $package = $this->packagistClient->get(static::PACKAGIST_GROUP.'/'.$project);
+        $this->io->title($package->getName());
+        $this->prepareRelease($package, $branch, $output);
+
+        return 0;
+    }
+
+    private function getProject(InputInterface $input, OutputInterface $output)
+    {
+        $helper = $this->getHelper('question');
+
+        $question = new Question('<info>Please enter the name of the project to release:</info> ');
+        $question->setAutocompleterValues(array_keys($this->configs['projects']));
+        $question->setNormalizer(static function ($answer) {
+            return $answer ? trim($answer) : '';
+        });
+        $question->setValidator(function ($answer) {
+            if (!\array_key_exists($answer, $this->configs['projects'])) {
+                throw new \RuntimeException('The name of the project should be on `projects.yml`');
+            }
+
+            return $answer;
+        });
+        $question->setMaxAttempts(3);
+
+        return $helper->ask($input, $output, $question);
+    }
+
+    private function prepareRelease(Package $package, $branch, OutputInterface $output): void
+    {
+        $repositoryName = $this->getRepositoryName($package);
+
+        $currentRelease = $this->githubClient->repo()->releases()->latest(
+            static::GITHUB_GROUP,
+            $repositoryName
+        );
+
+        $branchToRelease = $this->githubClient->repo()->branches(
+            static::GITHUB_GROUP,
+            $repositoryName,
+            $branch
+        );
+
+        $statuses = $this->githubClient->repo()->statuses()->combined(
+            static::GITHUB_GROUP,
+            $repositoryName,
+            $branchToRelease['commit']['sha']
+        );
+
+        $pulls = $this->findPullRequestsSince($currentRelease['published_at'], $repositoryName, $branch);
+        $nextVersion = $this->determineNextVersion($currentRelease['tag_name'], $pulls);
+        $changelog = array_reduce(
+            array_filter(array_column($pulls, 'changelog')),
+            'array_merge_recursive',
+            []
+        );
+
+        $this->io->section('Project');
+
+        foreach ($statuses['statuses'] as $status) {
+            $print = $status['description']."\n".$status['target_url'];
+
+            if ('success' === $status['state']) {
+                $this->io->success($print);
+            } elseif ('pending' === $status['state']) {
+                $this->io->warning($print);
+            } else {
+                $this->io->error($print);
+            }
+        }
+
+        $this->io->section('Pull requests');
+
+        foreach ($pulls as $pull) {
+            $this->printPullRequest($pull, $output);
+        }
+
+        $this->io->section('Release');
+
+        if ($nextVersion === $currentRelease['tag_name']) {
+            $this->io->warning('Release is not needed');
+        } else {
+            $this->io->success('Next release will be: '.$nextVersion);
+
+            $this->io->section('Changelog');
+
+            $this->printRelease($currentRelease['tag_name'], $nextVersion, $package, $output);
+            $this->printChangelog($changelog, $output);
+        }
+    }
+
+    private function printPullRequest($pull, OutputInterface $output): void
+    {
+        if (\array_key_exists($pull['stability'], static::$stabilities)) {
+            $output->write('<fg=black;bg='.static::$stabilities[$pull['stability']].'>['
+                .strtoupper($pull['stability']).']</> ');
+        } else {
+            $output->write('<error>[NOT SET]</error> ');
+        }
+        $output->write('<info>'.$pull['title'].'</info>');
+
+        foreach ($pull['labels'] as $label) {
+            if (!\array_key_exists($label['name'], static::$labels)) {
+                $output->write(' <error>['.$label['name'].']</error>');
+            } else {
+                $output->write(' <fg='.static::$labels[$label['name']].'>['.$label['name'].']</>');
+            }
+        }
+
+        if (empty($pull['labels'])) {
+            $output->write(' <fg=black;bg=yellow>[No labels]</>');
+        }
+
+        if (!$pull['changelog'] && 'pedantic' !== $pull['stability']) {
+            $output->write(' <error>[Changelog not found]</error>');
+        } elseif (!$pull['changelog']) {
+            $output->write(' <fg=black;bg=green>[Changelog not found]</>');
+        } elseif ($pull['changelog'] && 'pedantic' === $pull['stability']) {
+            $output->write(' <fg=black;bg=yellow>[Changelog found]</>');
+        } else {
+            $output->write(' <fg=black;bg=green>[Changelog found]</>');
+        }
+        $this->io->newLine();
+        $output->writeln($pull['html_url']);
+        $this->io->newLine();
+    }
+
+    private function printRelease($currentVersion, $nextVersion, Package $package, OutputInterface $output): void
+    {
+        $output->writeln('## ['.$nextVersion.']('
+            .$package->getRepository().'/compare/'.$currentVersion.'...'.$nextVersion
+            .') - '.date('Y-m-d'));
+    }
+
+    private function printChangelog($changelog, OutputInterface $output): void
+    {
+        ksort($changelog);
+        foreach ($changelog as $type => $changes) {
+            if (0 === \count($changes)) {
+                continue;
+            }
+
+            $output->writeln('### '.$type);
+
+            foreach ($changes as $change) {
+                $output->writeln($change);
+            }
+            $this->io->newLine();
+        }
+    }
+
+    private function parseChangelog($pull)
+    {
+        $changelog = [];
+        $body = preg_replace('/<!--(.*)-->/Uis', '', $pull['body']);
+        preg_match('/## Changelog.*```\s*markdown\s*\\n(.*)\\n```/Uis', $body, $matches);
+
+        if (2 == \count($matches)) {
+            $lines = explode(PHP_EOL, $matches[1]);
+
+            $section = '';
+            foreach ($lines as $line) {
+                $line = trim($line);
+
+                if (empty($line)) {
+                    continue;
+                }
+
+                if (0 === strpos($line, '#')) {
+                    $section = preg_replace('/^#* /i', '', $line);
+                } elseif (!empty($section)) {
+                    $line = preg_replace('/^- /i', '', $line);
+                    $changelog[$section][] = '- [[#'.$pull['number'].']('.$pull['html_url'].')] '.
+                        ucfirst($line).' ([@'.$pull['user']['login'].']('.$pull['user']['html_url'].'))';
+                }
+            }
+        }
+
+        return $changelog;
+    }
+
+    private function determineNextVersion($currentVersion, $pulls)
+    {
+        $stabilities = array_column($pulls, 'stability');
+        $parts = explode('.', $currentVersion);
+
+        if (\in_array('minor', $stabilities)) {
+            return implode('.', [$parts[0], $parts[1] + 1, 0]);
+        } elseif (\in_array('patch', $stabilities)) {
+            return implode('.', [$parts[0], $parts[1], $parts[2] + 1]);
+        }
+
+        return $currentVersion;
+    }
+
+    private function determinePullRequestStability($pull)
+    {
+        $labels = array_column($pull['labels'], 'name');
+
+        if (\in_array('minor', $labels)) {
+            return 'minor';
+        } elseif (\in_array('patch', $labels)) {
+            return 'patch';
+        } elseif (array_intersect(['docs', 'pedantic'], $labels)) {
+            return 'pedantic';
+        }
+    }
+
+    private function findPullRequestsSince($date, $repositoryName, $branch)
+    {
+        $pulls = $this->githubPaginator->fetchAll($this->githubClient->search(), 'issues', [
+            'repo:'.static::GITHUB_GROUP.'/'.$repositoryName.
+            ' type:pr is:merged base:'.$branch.' merged:>'.$date,
+        ]);
+
+        $filteredPulls = [];
+        foreach ($pulls as $pull) {
+            if ('SonataCI' === $pull['user']['login']) {
+                continue;
+            }
+
+            $pull['changelog'] = $this->parseChangelog($pull);
+            $pull['stability'] = $this->determinePullRequestStability($pull);
+
+            $filteredPulls[] = $pull;
+        }
+
+        return $filteredPulls;
+    }
+}


jordisala1991:
I know the code is totally crap right now.

This is some example of output when executing this command against `doctrine-orm-admin-bundle`:

```

sonata-project/doctrine-orm-admin-bundle
========================================

Pull requests
-------------

added captions to TOC tree in docs (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/669
update one to many documentation (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/667
Fixed wrong translation in one_to_many template (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/666
Small clanup (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/662
Allow simplethings/entity-audit-bundle ^1.0 (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/661
Fix check access to object in view (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/660
Fixed duplicate translation of "Delete" in edit tab view (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/658
Translation is broken in oneToMany edit form (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/657
Fixed wrong entity link in list view (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/654
Use the right array conversion for filter value (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/653
Remove deprecated translation logic (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/651
TYPE_IS_NOT_EQUAL doesn't consider NULL values (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/650
TYPE_NOT_CONTAINS doesn't consider NULL values (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/649
Fixed deprecated BaseBlockService usage (#641) (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/648
Apply fixes from StyleCI (patch)
    https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/647


Changelog
---------

### Changed
- translation in twig templates uses the twig translation filter (#651)
### Fixed
- Fixed wrong translation in delete checkbox in `edit_orm_one_to_many_inline_table.html.twig` (#666)
- Fixed duplicate translation of "Delete" in edit tab view (#658)
- Fixed broken translation in oneToMany table view (#657)
- Fixed wrong link generation when calling `DatagridMapper::addIdentifier` on mapped field (#654)
- Use the right array conversion for filter value (#653)
- Consider NULL values when using 'is not equal' advanced model filter (#650)
- Consider NULL values when using 'does not contain' advanced string filter (#649)
- deprecated usage of Sonata\BlockBundle\Block\BaseBlockService (#648)
### Security
- Fixed view - check specific item collection, not to the whole collection. (#660)
```

src/Console/Command/ReleaseCommand.php
```diff
@@ -0,0 +1,174 @@
...
+    {
+        $project = $input->getArgument('project');
+        $branch = $input->getOption('branch');
+
+        try {
+            $package = $this->packagistClient->get(static::PACKAGIST_GROUP.'/'.$project);
+            $this->io->title($package->getName());
+            $this->prepareRelease($package, $branch, $output);
+        } catch (ExceptionInterface $e) {
+            $this->io->error('Failed with message: '.$e->getMessage());
```
greg0ire: What's the point? Also, maybe not return 0?

jordisala1991: Just copied that part from another command of dev-kit. Not sure if it is needed :D

jordisala1991:
Disabled travis for now on this PR, I don't want to trigger a lot of builds when pushing commits

sstok:
https://github.com/park-manager/hubkit Feel free to reuse the Version class which automatically determines the next possible version, validates gaps, etc 😃

jordisala1991:
Hmm, that would be nice, but I know what the next version should be by looking at the labels of all PRs. Is there a way to do that with your class @sstok ?

jordisala1991:
I have a question about github.

When you tag a new version, how do you mark it as the latest release?

The thing is, when executing this command, I check for the latest release, but looking at `doctrine-orm-admin-bundle` the latest release is not the latest :trollface: .

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/releases

jordisala1991:
We can use this: https://developer.github.com/v3/repos/contents/#get-contents

To get the content of the changelog, then prepend the new changelog and use that other method:

https://developer.github.com/v3/repos/contents/#update-a-file

To update it and create a commit for the next release, then use this:

https://developer.github.com/v3/git/tags/#create-a-tag-object

To create the tag pointing to that commit sha, and finally use this:

https://developer.github.com/v3/repos/releases/#create-a-release

To create the release pointing to that tag. 🎉 

WDYT @sonata-project/contributors, sounds nice, but also risky if something goes wrong

sstok:
> I mean, given a current release and a minor, patch string or integer give me the next release

Yes, you can take a version say v1.2.4 and ask for an increase of any part, including alpha, beta, rc, etc. v1.2.4 + minor = v1.3.0 

https://github.com/park-manager/hubkit/blob/master/src/Helper/Version.php#L177

But there is a minor catch if you want to jump from v1.2.4 to 2.0.0-ALPHA1, you need to create this version explicit as you can also have v1.3.0-ALPHA1.

And all versions must follow the semver scheme (`[v]major.minor.patch[-mataver]`), nightly builds are simple ignored.

I will move these classes to there own package today (or tomorrow 😉 ).

jordisala1991:
Thanks @sstok , will try to refactor the `determineNextVersion` method using your package. 👍

src/Console/Command/ReleaseCommand.php
```diff
@@ -0,0 +1,289 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Jordi Sala <jordism91@gmail.com>
+ */
+final class ReleaseCommand extends AbstractCommand
+{
+    protected static $labels = array(
```
greg0ire: Protected property on a final class?

jordisala1991: 👍 changed to private

src/Console/Command/ReleaseCommand.php
```diff
@@ -0,0 +1,289 @@
...
+        'bug' => 'red',
+        'docs' => 'yellow',
+        'minor' => 'green',
+        'pedantic' => 'cyan',
+    );
+
+    private static $stabilities = array(
+        'patch' => 'blue',
+        'minor' => 'green',
+        'stable' => 'yellow',
```
OskarStark: shouldn't this be `major`

src/Console/Command/ReleaseCommand.php
```diff
@@ -0,0 +1,289 @@
...
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        parent::configure();
+
+        $this
+            ->setName('release')
+            ->setDescription('Makes a release of a project.')
+            ->addArgument('project', InputArgument::REQUIRED, 'The project to do a release of.')
```
OskarStark: could you please add a help command with examples?

src/Console/Command/ReleaseCommand.php
jordisala1991: I use stable for pedantic changes. I didn't add major because major release will require special work, that maybe cannot be automated, or atleast not in the same way

src/Console/Command/ReleaseCommand.php
jordisala1991: Yes, do you mean a comment with some examples or the class command itself has a method to do that?

OskarStark: screenshot 2017-03-20 12 22 20

Something like this

src/Console/Command/ReleaseCommand.php
OskarStark: ok

greg0ire: What's wrong with `pedantic`? `stable` is a bit vague IMO, and `patch` / `minor` releases could happen on a stable branch, so it might be confusing.

jordisala1991: I would like a word for PRs that does not require a release. We could use pedantic, but then docs PR should get this pedantic status.

Summary:

Given a currentRelease 3.3.0 and a Pull request with the following status:

'patch' => will result in 3.3.1
'minor' => will result in 3.4.0
'????' => will result in 3.3.0 (no release needed)
'major' => will result in 4.0.0 (not the scope of this PR)

What word should I use? @greg0ire @OskarStark

greg0ire: Harmless? No_impact? Irrelevant? Idem potent?

jordisala1991: Like irrelevant 👍

greg0ire: Unnoticeable? Furtive?

greg0ire: Lol

src/Console/Command/ReleaseCommand.php
jordisala1991: Added, my english is pretty bad, but @greg0ire could help improving the text :P .

src/Console/Command/ReleaseCommand.php
```diff
@@ -0,0 +1,304 @@
...
+        $help = <<<'EOT'
+The <info>release</info> command analyzes pull request of a given project to determine
+the changelog and the next version to release.
+
+Usage:
+
+<info>php dev-kit release doctrine-orm-admin-bundle</info>
+
+The command will show what is the status of the project, then a list of pull requests
+made against selected branch (default: <info>3.x</info>) with the following information: stability, name,
```
OskarStark: couldn't we parse our projects.yml and mark the default branch as default there?

src/Console/Command/ReleaseCommand.php
```diff
@@ -0,0 +1,304 @@
...
+labels, changelog, url.
+
+After that, it will show what is the next version to release and the changelog for that release.
+EOT;
+
+        $this
+            ->setName('release')
+            ->setDescription('Helps with a project release.')
+            ->addArgument('project', InputArgument::REQUIRED, 'The project to release.')
```
OskarStark: what about making this also interactive and having autocomplete byes on the `projects.yml` ❓

src/Console/Command/ReleaseCommand.php
```diff
@@ -0,0 +1,304 @@
...
+            ->addOption('branch', null, InputOption::VALUE_OPTIONAL, 'branch to release', '3.x')
```
OskarStark: same autocompletion here?

jordisala1991: I Removed this parameter, not the branch detection is done based on projects.yml

src/Console/Command/ReleaseCommand.php
jordisala1991: Changed this copy, see comment below

src/Console/Command/ReleaseCommand.php
jordisala1991: Implemented

jordisala1991:
My proposal is to leave this PR with only a command to show what the current state of the branch and what will be the changelog of the new version, status of merged PRs, etc..

And create another PR after this one is merged to do the automatic Pull request with the new changelog (maybe with automerge) and show to the user what commands needs to execute to create the tag (maybe?)

WDYT @sonata-project/contributors

OskarStark:
I executed it for the `translation-bundle` and got this result:
screenshot 2017-03-22 10 16 57

totally fine, except one thing:
screenshot 2017-03-22 10 16 18

This is **not** `IRRELEVANT`, there is only the `require review` label. We should verify, that at least one of `docs, pedantic, patch, minor` is assigned, I would leave `major` out, because this tool should only help create minor and patch versions for now, right?

In case there is not one of these labels applied, we should point this out, that the user of this command can fix this PR

OskarStark:
And currently I cannot execute this command for `ecommerce`

OskarStark:
as we were generating this changelog from now on, should we add the authors like symfony does?

ping @sonata-project/contributors

greg0ire:
The authors would be a really really nice touch!

jordisala1991:
Give me the format of authors, and I will add it.

jordisala1991:
I already added The pull request associated with every line of the changelog. But agree that having the author is nice too

jordisala1991:
> This is not IRRELEVANT, there is only the require review label. We should verify, that at least one of docs, pedantic, patch, minor is assigned, I would leave major out, because this tool should only help create minor and patch versions for now, right?

Those red labels are not suposed to be there, but I agree that is will be nice to have an alert about not having any of those labels.

greg0ire:
`- my change (@author)` would be best IMO (so, using the github handle, `@` included, not sure why sf does not include it)

jordisala1991:
https://github.com/symfony/symfony/blob/master/CHANGELOG-3.0.md

Not sure if github handles `@` , because it does not handle `#`

greg0ire:
Hum right… that's weird…

OskarStark:
What about separating the stuff like this

## `CHANGELOG.md`
```
## [3.1.3](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/compare/3.1.2...3.1.3) - 2017-03-22
### Fixed
- [#656] Fixed foo (OskarStark)
- [#655] Fixed bar (greg0ire)
- [#654] Fixed baz (jordisala1991)
```

## For the release tab:
Meaning this one: https://github.com/sonata-project/SonataAdminBundle/releases

This should respect the user handle
```
## [3.1.3](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/compare/3.1.2...3.1.3) - 2017-03-22
### Fixed
- [#656](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/656) Fixed foo (OskarStark)
- [#655](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/655) Fixed bar (greg0ire)
- [#654](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/654) Fixed baz (jordisala1991)
```

Imo it is more convenient to put the ticket number in front of the text, but it really doesn't matter

jordisala1991:
Implemented new changelog format.

Aaaand implemented a message when a PR does not contain any stability labels

jordisala1991:
captura de pantalla 2017-03-22 a las 20 39 16

jordisala1991:
> And currently I cannot execute this command for ecommerce

@OskarStark it does not work with ecommerce because ecommerce has only 3 releases, and none of them is marked as latest release, so github method: `getLatestRelease` returns an exception. IMO this should be fixed on that repo, if 2.0.0 is the latest release, just mark it.

I don't know if only @Soullivaneuh can do that, or we can do it also

jordisala1991:
I think we can do it also, shall we do it? @greg0ire

OskarStark:
Yes, please do it @jordisala1991 

btw. good catch 😄

jordisala1991:
IMO we should catch maybe if the lastest tag is not the latest release.

https://github.com/sonata-project/dev-kit/pull/204/files#diff-55aa24e78df8010aea154e9506cab16dR361

If this is correct I will replace irrelevant by pedantic. ok?

jordisala1991:
And I have to restore project argument but in an optional way, because maybe we want to run this command each week and post results on slack programatically

OskarStark:
> If this is correct I will replace irrelevant by pedantic. ok?

ok

> And I have to restore project argument but in an optional way

great idea

And maybe we should add another command, which checks all projects and show how many commits we are on top of the latest release, I think this would be more useful information in slack

jordisala1991:
Indeed, then maybe dint add project argument back?

OskarStark:
Yes, why not!

jordisala1991:
YAGNI reason :trollface:

core23:
Maybe we could add an auto-tweet to this command to?

jordisala1991:
> Maybe we could add an auto-tweet to this command to?

That could be nice but for the second iteration of this command. Let's keep this PR for just the information about the release: Pull requests, changelog, release number, build status...

jordisala1991:
I changed the way we retrieve pull request, not it is more efficient, and only retrieve those pull requests already merged and > currentRelease date.

DashboardBundle fails because for some reason github returns a different kind of object when finding release >.<

jordisala1991:
Fixed, it was because there is only one branch on DashboardBundle, I changed a bit the code to check if there is only one branch.

jordisala1991:
> For future problems we should catch the exception and point out, that no release is marked as latest release an where the maintainer can do this, WDYT?

Implemented. Maybe we should remove it once all project have an stable release.

> IMO we should catch maybe if the lastest tag is not the latest release.

Super hard to implement, or atleast in a good way. Let's ignore this for now... it should not happen

jordisala1991:
@OskarStark found a bug related to parsing `changelog`. That's why I improved that part.

Now the regex will be able to find the changelog even if it is not using ### for titles, or if there are whitespaces between ``` and `markdown` or after it.

The only requirements are, `## Changelog` and a markdown section with an ending (```).

jordisala1991:
Not sure if i have to change something, or it is working ATM and waiting for @Soullivaneuh review

OskarStark:
@Soullivaneuh can you please merge this?

OskarStark:
IMO this PR was a big amount of work and @jordisala1991 did it great!

This PR was opened 3 Months ago and still no merge.
@Soullivaneuh what is still todo?
@jordisala1991 are you willing to write tests or could we do this in a separate step.

It would be great to get a tool which provides as much as it can help to create a new release...

jordisala1991:
I can write some tests, but I will also like to know the current state of this PR first. (And IMO it can be done in another step, because this will delay a lot this PR, it won't be easy to test everything)

OskarStark:
We should merge this

jordisala1991:
Soon this will be a year old, 🎉

greg0ire:
That's cute! https://github.com/tpope/vim-pathogen/pull/142

src/Console/Command/ReleaseCommand.php
```diff
@@ -0,0 +1,318 @@
...
+    );
+
+    private static $stabilities = array(
+        'patch' => 'blue',
+        'minor' => 'green',
+        'pedantic' => 'yellow',
+    );
+
+    /**
+     * {@inheritdoc}
```
kunicmarko20: Nice comment :trollface:

jordisala1991: Don't worry, I will die before this gets merged :trollface:

src/Console/Command/ReleaseCommand.php
```diff
@@ -0,0 +1,318 @@
...
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $project = $this->getProject($input, $output);
+        $branches = array_keys($this->configs['projects'][$project]['branches']);
+        $branch = count($branches) > 1 ? next($branches) : current($branches);
+
+        try {
+            $package = $this->packagistClient->get(static::PACKAGIST_GROUP.'/'.$project);
```
kunicmarko20: why `static::` and not self?

core23:
Can we finally merge this, if this is still an open topic...

src/Console/Command/ReleaseCommand.php
greg0ire: I think Doctrine uses "Improvement". It's nice because it does not feel insulting for the contributor.

greg0ire:
> Note: No request change but I will handle the rest manually later.

Does somebody know what this means?

jordisala1991:
No idea, does the PR is still useful?

greg0ire:
I will try it for next release and see how it compares to the bash script I wrote, maybe we will end up with a mix of both :P

src/Console/Command/ReleaseCommand.php
```diff
@@ -0,0 +1,318 @@
...
+    {
+        $repositoryName = $this->getRepositoryName($package);
+
+        try {
+            $currentRelease = $this->githubClient->repo()->releases()->latest(
+                static::GITHUB_GROUP,
+                $repositoryName
+            );
+        } catch (ExceptionInterface $e) {
+            throw new RuntimeException('Current release not found');
```
greg0ire, line 123:
```suggestion
            throw new RuntimeException('Current release not found', 0, $e);
```

src/Console/Command/ReleaseCommand.php
greg0ire: Let's remove the try catch, it just makes things like bad credentials harder to troubleshoot IMO

src/Console/Command/ReleaseCommand.php
jordisala1991: I removed this exception to help debuggin problems, just like the other one

jordisala1991:
I rebased the PR, squashed all commits, and fixed some of the reviews you have made here.
I can't merge your PR @greg0ire it will need a rebase too. A lot of things changed here.

For the command... IMO we should go with a similar strategy than Symfony uses. They have several commands, but the most important ones are (seen in a Symfony meetup): 

1. Prepare the release: it compiles all the changelog and automatically creates a PR with it.
2. Release: once that PR is merged they have another command that automatically creates the tag, creates the release and attach the changelog.

This PR is going for the first step, we still missing the automatic PR creation but it can be done later too. Once we are good for the first step, we can go for the second to automate the process of creating the tag and the release

greg0ire:
I agree, let's not try to do everything at once.

greg0ire:
Thanks @jordisala1991 !

jordisala1991:
Thank you :) I will start thinking of next steps

PR #245, opened 2017-03-27
removed obsolete labels

OskarStark:
removed "review required" and "RTM" label, as github now provides a better workflow to handle PR's

### DIFF
--- a/config/dev-kit.yml
+++ b/config/dev-kit.yml
@@ -23,12 +23,8 @@ labels:
     color: 'ededed'
   - name: in progress
     color: 'ededed'
-  - name: RTM
-    color: 'ffffff'
   - name: docs
     color: 'fbca04'
-  - name: review required
-    color: 'd4c5f9'
   - name: help wanted
     color: '0e8a16'
   - name: Easy Pick


OskarStark:
Maybe keep the `RTM` only for dashboard as @Soullivaneuh relies on this label I think

jordisala1991:
Remove this labels will require another PR following this one to remove some code from here: https://github.com/sonata-project/dev-kit/blob/master/src/GithubHookProcessor.php#L72

Soullivaneuh:
> as github now provides a better workflow to handle PR's

Could you please elaborate @OskarStark?

As @jordisala1991 says, this label is used for automation and easy filtering, how could we replace it?

Soullivaneuh:
> Maybe keep the RTM only for dashboard as @Soullivaneuh relies on this label I think

Same thing here. Do approvals really replace the fact the MR is ready to merge?

OskarStark:
> Same thing here. Do approvals really replace the fact the MR is ready to merge?

Ok fine lets keep them, and filtering is great, but "review required" should then be removed if the PR is closed and merged, otherwise its useless, isn't it?

PR #247, opened 2017-06-05
Branch management: the 3 roads

### LINKED ISSUE #153: Branch management: The three roads
Currently, we manage two branches: The stable one and the master one.

This works great, but this will make big problems when we will start releasing new majors.

Concrete case with admin-bundle and core-bundle:

If we release core-bundle right now, the logic would says that 3.x is not maintained anymore.

But, admin-bundle wont be compatible with core-bundle 4.0 right now. So user will still use the legacy branch of core-bundle waiting compatibility.

But what if a bug is found on core-bundle 3.x? Should we refuse the fix? I don't think so.

This is why I propose to maintains 3 branches starting the next major:
- `master`: The unstable branch, like actual (5.x for core-bundle)
- `stable`: For BC features and deprecation (4.x for core-bundle)
- `legacy`: For bug fixes. All bug-fixes should be submitted here except those concerning a feature added on the `stable` branch.

Any suggestions?

The dispatcher has to be updated before releasing new majors.

Soullivaneuh:
Closes #153

### DIFF
--- a/project/CONTRIBUTING.md
+++ b/project/CONTRIBUTING.md
@@ -162,9 +162,9 @@ Notes:
   * Branch `{{ legacy_branch }}` is the branch of the **latest stable** patch releases and
   has to be used for Backward Compatible bug fixes only.
   * Branch `{{ stable_branch }}` is the branch of the **latest stable** minor release and
-  has to be used for Backward compatible PRs.
+  has to be used for Backward compatible enhancement PRs.
   **No bug fix will be accepted here**, except if the bug fix concerns `{{ stable_branch }}` and **only this one**.
-  Bug fixes merged on `{{ legacy_branch }}` will be reported on other branches by fallback merging.
+  Bug fixes merged on `{{ legacy_branch }}` will be ported on other branches by fallback merging.
   * If you PR is not **Backward Compatible** but can be, it **must** be:
     * Changing a function/method signature? Prefer create a new one and deprecate the old one.
     * Code deletion? Don't. Please deprecate it instead.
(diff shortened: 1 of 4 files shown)


project/CONTRIBUTING.md
```diff
@@ -150,16 +150,21 @@ Here is a short table resuming on which you have to start:
...
 Feature              | Yes                      | Minor           | `{{ stable_branch }}`   | |
 Feature              | No (Only if no choice)   | Major           | `{{ unstable_branch }}` | |
 Deprecation          | Yes (Have to)            | Minor           | `{{ stable_branch }}`   | |
 Deprecation removal  | No (Can't be)            | Major           | `{{ unstable_branch }}` | |
 
 Notes:
+  * Branch `{{ legacy_branch }}` is the branch of the **latest stable** patch releases and
+  has to be used for Backward Compatible bug fixes only.
   * Branch `{{ stable_branch }}` is the branch of the **latest stable** minor release and
   has to be used for Backward compatible PRs.
```
greg0ire: "Backward compatible PRs" => "Backward compatible enhancement PRs"?

project/CONTRIBUTING.md
```diff
@@ -150,16 +150,21 @@ Here is a short table resuming on which you have to start:
...
+  **No bug fix will be accepted here**, except if the bug fix concerns `{{ stable_branch }}` and **only this one**.
+  Bug fixes merged on `{{ legacy_branch }}` will be reported on other branches by fallback merging.
```
greg0ire: "reported" => "ported" or "backported"

project/CONTRIBUTING.md
Soullivaneuh: Yeah, it could be great to differentiate this.

PR #248, opened 2017-06-14
StyleCI: Ignore assets from vendors

core23:
We shouldn't fix assets from vendors when running StyleCI.

Refs. https://github.com/sonata-project/SonataFormatterBundle/pull/255


Soullivaneuh:
The best would be to not have vendor... but I guess we don't have the choice now.

@core23 Please provide a working testing PR of your modification.

core23:
@Soullivaneuh Have a look at https://github.com/sonata-project/SonataFormatterBundle/pull/257

PR #249, opened 2017-06-30
Drop support for php 7.0 on next major

greg0ire:


config/projects.yml
```diff
@@ -182,7 +182,7 @@ doctrine-phpcr-admin-bundle:
   branches:
     master:
       # sonata 2 should still support php 5.6 to be consistent with the symfony cmf
-      php: ['5.6', '7.0', '7.1']
+      php: ['5.6', '7.1']
```
greg0ire: @dbu @lsmith77 please confirm you are ok with this

config/projects.yml
```diff
@@ -257,7 +257,7 @@ google-authenticator:
   docs_target: false
   branches:
     master:
-      php: ['7.0', '7.1']
+      php: ['7.1']
     2.x:
       php: ['7.0', '7.1']
```
greg0ire: Too late it is released

config/projects.yml
dbu: we are about to release 2.0, so i think its too early to drop 7.0. also dropping 7.0 but not 5.6 would be rather pointless. i suggest we wait until we tagged the stable release and created a 2.x branch and only then drop 5.6 and 7.0 for master that will be 3.x

greg0ire: Ok I'll revert this change :+1:

greg0ire:
> what is the time horizon of the next majors? be aware that according to http://php.net/supported-versions.php php 7.0 is still actively supported until end of this year and security fixes until next year...

Since there is a lot of inter-dependencies, and a lot of package, the releases will not be synchronized (I think it is for the better). For instance, google authenticator 2 is already released, which, by the way, means there already is an exception to the "drop php 7.0" proposal

greg0ire:
> be aware that according to http://php.net/supported-versions.php php 7.0 is still actively supported until end of this year and security fixes until next year...

Since it's a next major, I personally think we should start by supporting only the latest versions of our deps, especially this one. We're not hurting anyone by doing that, and since we release at a very slow pace (at least that's how I see it), we shouldn't choose to get stuck with 7.0 for a potentially long period of time.

Soullivaneuh:
Rest in Peace, PHP 7.0.

PR #251, opened 2017-06-30
Comment out piece of advice

greg0ire:
We don't actually want to display this.

### DIFF
--- a/project/.github/PULL_REQUEST_TEMPLATE.md
+++ b/project/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,7 +9,7 @@
 -->
 I am targeting this branch, because {reason}.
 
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
I disagree with this. I voluntarily made it visible to remember contributors **and** mergers to take car of this, especially for 3 branches repositories.

If you have a better solution, please tell, but we can't just hide it.

greg0ire:
A bot that would comment when labelling with "patch" on the stable branch would be a better solution IMO, but it's of course more work.

Soullivaneuh:
Maybe, only if patch label are correctly used. :trollface: 

Waiting that, if we don't have another solution, let's keep this comment.

PR #254, opened 2017-07-08
Remove deprecated versions of sonata_user from SonataIntlBundle

jordisala1991:
We are building IntlBundle with SonataUser 2.* which is deprecated and on the conflict section of composer.json.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -269,12 +269,12 @@ intl-bundle:
       php: ['7.1']
       versions:
         symfony: ['2.8', '3.1', '3.2']
-        sonata_user: ['2', '3']
+        sonata_user: ['3']
     2.x:
       php: ['5.3', '5.4', '5.5', '5.6', '7.0', '7.1']
       versions:
         symfony: ['2.3', '2.7', '2.8', '3.1', '3.2']
-        sonata_user: ['2', '3']
+        sonata_user: ['3']
 
 media-bundle:
   branches:


config/projects.yml
```diff
@@ -269,12 +269,12 @@ intl-bundle:
       php: ['7.1']
       versions:
         symfony: ['2.8', '3.1', '3.2']
-        sonata_user: ['2', '3']
+        sonata_user: ['3']
     2.x:
       php: ['5.3', '5.4', '5.5', '5.6', '7.0', '7.1']
       versions:
         symfony: ['2.3', '2.7', '2.8', '3.1', '3.2']
-        sonata_user: ['2', '3']
```
Soullivaneuh: This should not be removed from current stable: https://github.com/sonata-project/SonataIntlBundle/blob/2.3.1/composer.json#L37

It would be BC break.

Please rever this line.

core23: We don't support 2.x anymore. IMHO we should risk this break. It looks like a mistake we've done in the past.

Soullivaneuh: @core23 Could you please elaborate?

The 2.x version was included on stable releases of this bundle, so it is supported, right?

This can be removed on master, not here IMHO.

greg0ire: We can safely drop support for unmaintained software IMO. Didn't we already discuss this somewhere BTW? Rings a bell...

jordisala1991:
Closing, #259 covers this case and more :)

PR #255, opened 2017-08-01
Drop outdated versions for dashboard bundle

core23:
We have no stable release yet for this bundle and should only support the latest supported versions for php and symfony.

Refs: https://github.com/sonata-project/SonataDashboardBundle/pull/82


Soullivaneuh:
@core23 We can't do this because the last LTS version is 2.8 and we have to support at least one LTS.

This may be done later when the next LTS (3.4 if I'm right) will be out.

PR #259, opened 2017-08-05
Stop running tests against Sonata 2.x

greg0ire:
We no longer support that.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -191,8 +191,8 @@ doctrine-phpcr-admin-bundle:
       php: ['5.3', '5.4', '5.5', '5.6', '7.0', '7.1']
       versions:
         symfony: ['2.3', '2.7', '2.8']
-        sonata_admin: ['2']
-        sonata_block: ['2']
+        sonata_admin: ['3']
+        sonata_block: ['3']
 
 easy-extends-bundle:
   branches:
@@ -269,12 +269,12 @@ intl-bundle:
       php: ['7.1']
       versions:
         symfony: ['2.8', '3.1', '3.2']
-        sonata_user: ['2', '3']
+        sonata_user: ['3']
     2.x:
       php: ['5.3', '5.4', '5.5', '5.6', '7.0', '7.1']
       versions:
         symfony: ['2.3', '2.7', '2.8', '3.1', '3.2']
-        sonata_user: ['2', '3']
+        sonata_user: ['3']
 
 media-bundle:
   branches:


greg0ire:
I'm going ahead and doing it manually: https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/482, to keep the ball rolling. dev-kit should try to put it back tonight if you don't merge this, so please hurry.

config/projects.yml
```diff
@@ -269,12 +269,12 @@ intl-bundle:
       php: ['7.1']
       versions:
         symfony: ['2.8', '3.1', '3.2']
-        sonata_user: ['2', '3']
+        sonata_user: ['3']
     2.x:
       php: ['5.3', '5.4', '5.5', '5.6', '7.0', '7.1']
       versions:
         symfony: ['2.3', '2.7', '2.8', '3.1', '3.2']
-        sonata_user: ['2', '3']
```
Soullivaneuh: Not on stable branches.

config/projects.yml
```diff
@@ -191,8 +191,8 @@ doctrine-phpcr-admin-bundle:
       php: ['5.3', '5.4', '5.5', '5.6', '7.0', '7.1']
       versions:
         symfony: ['2.3', '2.7', '2.8']
-        sonata_admin: ['2']
```
Soullivaneuh: Not on stable branches.

jordisala1991: https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/1.x/composer.json

1.x does not allow SonataAdmin 3.x versions. So this should not be changed.

config/projects.yml
jordisala1991: Maybe we should leave this unchanged, just to not break people using 2.x of SonataUser with 2.x of IntBundle?

IMO if we can't change composer.json (this line: https://github.com/sonata-project/SonataIntlBundle/blob/2.x/composer.json#L37) there is no reason to change this.

It doesn't matter a lot because it is only used here: https://github.com/sonata-project/SonataIntlBundle/blob/2.x/Timezone/UserBasedTimezoneDetector.php

So, if it is on the same namespace and it has `getTimezone()` function, it will "work", or atleast not break on any version.

WDYT @greg0ire ?

greg0ire: > there is no reason to change this

I think there is a difference between allowing people to stay on 2.x and making sure tests pass for it. It means more maintenance for no good reason. IMO we should spend no time on it. Changing `composer.json` to explicitely disallow it is another discussion.

greg0ire: @Soullivaneuh @jordisala1991 if tomorrow, someone opens a ticket saying Intl no longer works with Sonata 2, won't we tell them to upgrade? IIRC we always close that type of tickets.

Soullivaneuh: > if tomorrow, someone opens a ticket saying Intl no longer works with Sonata 2, won't we tell them to upgrade? IIRC we always close that type of tickets.

The question is: Why this stable branch would not work with 2.x other bundle anymore? This could be done on next major like for SF upgrades.

Or it's a special case I didn't hear about. :-)

Soullivaneuh:
Discussed with @greg0ire on Slack. This is ok because last tag is a 2.x branch and the branch is failing for.. some reasons.

So this is a special case.

config/projects.yml
greg0ire: @jordisala1991 it can safely be removed

greg0ire: Ooooh I forgot I changed it instead of doing just a removal. No idea why the admin job builds though.

PR #260, opened 2017-08-07
Replace script with empty file

greg0ire:
dev-kit does not support deletions. Let remove the contents of the file
instead.


project/.travis/before_script_test.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env sh
+# legacy file, to be removed with dev-kit once we figure out how to do so
```
Soullivaneuh: No need of this comment. An issue could be better instead.

Issue #197, opened 2016-08-20
Using phpunit from the vendors


greg0ire:
@sonata-project/contributors, I have big issue with phpunit. The fact that we don't require and treat it as a dependency of our project seems to mean that as soon as we require a library the phpunit phar also bundles, test can fuck up in unexpected ways. It is the case in [this PR](https://github.com/sonata-project/SonataAdminBundle/pull/4029) since I required `matthiasnoback/symfony-dependency-injection-test` to test the Sonata Admin Symfony extension (which was not covered at all, it seems, by the way). My last commit, where I require and use phpunit provided by Composer fixes this issue. [Here is what I get otherwise](phpDocumentor/ReflectionDocBlock#80). This way of doing things has caused issues in the past, and will probably cause more if we continue like this. Can we agree that this is no longer relevant? As an argument against this, doing things like that also makes the travis workflow unnecessarily complicated IMO.
And finally, [we were warned against this by Jordi himself](https://github.com/sebastianbergmann/phpunit/issues/2014#issuecomment-166712695), and I really think we should listen to what this guy has to say.

Related issues (each one has been very time-consuming) : 
- https://github.com/matthiasnoback/SymfonyDependencyInjectionTest/issues/48
- https://github.com/sebastianbergmann/phpunit/issues/1757

core23:
Not sure if I understand the problem correctly... You want to include the composer depdency in the `composer.json` file? Sounds like a good idea, because we can define which version should be used for test classes (PHPunit 4 vs. 5). The latest 5.4 version also spams some deprecation warnings when running tests.

Adding it to composer would also improve autocompletion when writing new tests 👍

greg0ire:
> Not sure if I understand the problem correctly...

Yes you did!

Right now if I don't, phpunit can get confused when requiring a class. Should it autoload it from the phpunit phar, or from the project's vendors? This is why I can type `phpunit --version`, and have another version be displayed when running the tests.

core23:
> The main argument against using the version from the vendors is that you add some more constraints that can hold you back, because phpunit has many dependencies. 

Only interesting while development, but who cares? As long as a bundle user won't have to load it 👍

greg0ire:
> Only interesting while development, but who cares? As long as a bundle user won't have to load it :+1: 

Exactly. So on the one hand, we have hypothetical problems about being held back by too many deps, and on the other hand, we have the very real problem (happened to me personally several times) that we loose hours debugging the very strange problems this can create.

[...]

PR #265, opened 2017-08-16
WIP: Use simple-phpunit to run the tests

greg0ire:
We do not have to care anymore about picking the right phpunit versions,
and this should spare us some autoloading issues.
Closing #197

### DIFF
--- a/project/.travis/install_test.sh
+++ b/project/.travis/install_test.sh
@@ -3,15 +3,6 @@ set -ev
 
 mkdir --parents "${HOME}/bin"
 
-# PHPUnit install
-if [ ${TRAVIS_PHP_VERSION} '<' '5.6' ]; then
-    PHPUNIT_PHAR=phpunit-4.8.phar
-else
-    PHPUNIT_PHAR=phpunit-5.7.phar
-fi
-wget "https://phar.phpunit.de/${PHPUNIT_PHAR}" --output-document="${HOME}/bin/phpunit"
-chmod u+x "${HOME}/bin/phpunit"
-
 # Coveralls client install
 wget https://github.com/satooshi/php-coveralls/releases/download/v1.0.1/coveralls.phar --output-document="${HOME}/bin/coveralls"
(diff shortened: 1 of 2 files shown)


greg0ire:
From https://symfony.com/doc/current/components/phpunit_bridge.html :

> If you still need to use prophecy (but not symfony/yaml), then set the SYMFONY_PHPUNIT_REMOVE env var to symfony/yaml.

Maybe I need to amend this PR to change this env var, because we sometimes use prophecy. We might also use sf/yaml, this needs to be checked. Not sure how to manage that (globally, or on a repo-per-repo basis)

greg0ire:
grepping yaml yields nothing. Let's assume prophecy is required by default.

greg0ire:
Another question: can we use sf 2.8 and require sf/phpunit-bridge 3.2?

iltar:
Yes, they do not clash. Symfony doesn't force 2.x for the bridge and the bridge does not force 3.x.

greg0ire:
Got it to work on SonataAdminBundle, but it was really hard. I think I need to make other proof PRs for other bundles.

greg0ire:
more proof PRs:

- https://github.com/sonata-project/SonataCoreBundle/pull/428
- https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/723

as you can see, some work is needed library-side for it to work.

Soullivaneuh:
> as you can see, some work is needed library-side for it to work.

So, this is not yet ready?

greg0ire:
No, not yet. I have to make PRs for all the needed changes so that there is no need to change anything when this gets deployed. Will try to get that done soon.

jlamur:
IMO, you could add a note on `CONTRIBUTING.md` to explain how to run tests.

core23:
Do we still need this @greg0ire ?

greg0ire:
Maybe not… it's been a long time since we last had an autoload-clash related bug.

PR #267, opened 2017-08-19
Require precise

greg0ire:
The trusty Travis image cannot run php 5.3, because, you know, it is EOL
since 2014. This is a quick fix, but maybe we should amend our policy of
not dropping support for minor versions of php to include an exception
for versions that are no longer maintained.

https://github.com/travis-ci/travis-ci/issues/2963
https://secure.php.net/supported-versions.php

Proof PR: https://github.com/sonata-project/SonataAdminBundle/pull/4612

### DIFF
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -24,6 +24,7 @@ services:
 
 {% endif %}
 sudo: false
+dist: precise
 
 cache:
   pip: true


core23:
Will this fix the failing builds like https://travis-ci.org/sonata-project/SonataMediaBundle ?

greg0ire:
No it won't. This can fix only 5.3 builds that are starting to get flaky.

jordisala1991:
This is not applied yet to all projects right?

greg0ire:
Should be overnight I think

Issue #232, opened 2017-02-04
Drop outdated PHP versions on stable branch


core23:
I want to start a new discussion (3rd attempt....) about dropping outdated PHP versions on the stable branch, because we stumple opon some issues with newer third party dependencies, e.g. HTTPlug which requires at least PHP 5.5.

According to semver dropping a minor depenency is no BC break at all. Also the feedback on my [twitter poll](https://twitter.com/core23de/status/817044245018185728) showed a similar response about supported depencies.

Could we drop the minor PHP versions on the stable branch or can we just allow failures for outdated PHP versions.

Refs 
https://github.com/sonata-project/SonataAdminBundle/pull/4276#discussion_r96574850
https://github.com/sonata-project/SonataSeoBundle/pull/161#issuecomment-277433232

greg0ire:
I think we should amend our policy so that things that are no longer supported (no security fixes, or even no bugfixes) can be dropped on the stable branch.

core23:
Maybe start a new poll on twitter for that?

greg0ire:
The other poll you made about php was similar, I think it's safe to assume our userbase is ok with this.

Soullivaneuh:
Personal opinion: Go f*ck*ng ahead! I'm already on 7.0!

In a more general way, we can't consider php as a simple dependencies, because it's a system dependency IMHO.

Plus, we can't forgot we decide to maintain major branches only (2.x, 3.x etc...). So if we push, for example a v3.12 release with php `^5.6` instead of `^5.3` and we found a big on v3.11, we won't be able to deliver a bug fix on a php 5.3 compatible release.

And no, we can't manage minor branches like Symfony IMHO, it would be overkill. We are not a big company and we don't have time for that.

In conclusion, it could be interesting to find or ask why Symfony do maintain php constraint on minor version instead of lot of other projects, like doctrine for example.

I's a :-1: for me ATM, especially because of bugfixes releases. Waiting more answers/information.

cc @rande

Soullivaneuh:
> The other poll you made about php was similar, I think it's safe to assume our userbase is ok with this.

I disagree with that. IMHO it's a big difference between major and minor releases.

Soullivaneuh:
Addition : Even if we decide to do this, we can't drop a major version of a dependency on a minor release, even for PHP or composer.

greg0ire:
[...]

PR #280, opened 2017-08-28
Disable/Enable xDebug on PHP 5.3 to avoid breaking travis with slow builds

jordisala1991:
Composer on PHP 5.3 doesn't disable xdebug by itself.

https://getcomposer.org/doc/articles/troubleshooting.md#xdebug-impact-on-composer

This documentation doesn't talk about specific PHP versions, but we can look at travis.

We will see this on the same PR looking at the composer update command:

**This is a PHP 5.3 build**: https://travis-ci.org/sonata-project/SonataAdminBundle/jobs/268565708#L480
**This is a PHP 5.4 build**: https://travis-ci.org/sonata-project/SonataAdminBundle/jobs/268565709#L478

As you can see, from PHP 5.4 it will disable xDebug just fine, so this is only for PHP 5.3 version.

This PR adds a condition that will only be on the branches with 5.3 as one of the tested versions and will only be executed for PHP 5.3 so it is safe for newer versions that can rely on that feature of composer.

This can be removed when none of the branches (newer and older) rely on PHP 5.3

Why is this needed? Some older builds with lower-deps are failing because composer is taking so much time:

https://travis-ci.org/sonata-project/SonataEasyExtendsBundle/jobs/268563720

### DIFF
--- a/project/.travis/before_install_test.sh.twig
+++ b/project/.travis/before_install_test.sh.twig
@@ -6,6 +6,15 @@ if [ "${TRAVIS_PHP_VERSION}" != "hhvm" ]; then
     TRAVIS_INI_FILE="$PHP_INI_DIR/travis.ini"
     echo "memory_limit=3072M" >> "$TRAVIS_INI_FILE"
 
+    {% if '5.3' in php %}
+    if [ "$TRAVIS_PHP_VERSION" '<' '5.4' ]; then
+        XDEBUG_INI_FILE="$PHP_INI_DIR/xdebug.ini"
+        if [ -f  "$XDEBUG_INI_FILE" ]; then
+            mv "$XDEBUG_INI_FILE" /tmp
+        fi
+    fi
+    {% endif %}
+
     {% if 'mongodb' in services %}
     if [ "$TRAVIS_PHP_VERSION" '<' '7.0' ]; then
         echo "extension=mongo.so" >> "$TRAVIS_INI_FILE"
--- a/project/.travis/before_script_test.sh
+++ b/project/.travis/before_script_test.sh
@@ -1,2 +1,0 @@
-#!/usr/bin/env sh
-set -ev
--- a/project/.travis/before_script_test.sh.twig
+++ b/project/.travis/before_script_test.sh.twig
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+set -ev
+
+{% if '5.3' in php %}
+if [ "$TRAVIS_PHP_VERSION" != "hhvm" ] && [ "$TRAVIS_PHP_VERSION" '<' '5.4' ]; then
+    PHP_INI_DIR="$HOME/.phpenv/versions/$(phpenv version-name)/etc/conf.d/"
+    XDEBUG_INI_FILE="/tmp/xdebug.ini"
+    if [ -f  "$XDEBUG_INI_FILE" ]; then
+        mv "$XDEBUG_INI_FILE" "$PHP_INI_DIR"
+    fi
+fi
+{% endif %}


jordisala1991:
With this PR: https://github.com/sonata-project/SonataEasyExtendsBundle/pull/121

There are 2 commits, the first one is required to make the build pass in the first place (with and without this change)

The second commit applies THIS changes (the ones on this PR)

Before applying the second commit:

captura de pantalla 2017-08-28 a las 14 35 39

After applying this second commit:
captura de pantalla 2017-08-28 a las 14 24 48

As you can see, the difference is quite big (double the speed on composer update). I picked a bundle with not a lot of tests because I just wanted to know how much will improve the composer part.

jordisala1991:
Why do you think that @greg0ire ? With trusty we couldn't even make PHP5.3 work because it was not supported.

greg0ire:
I forced précise for 5.3 jobs recently. All the other jobs now use trusty

greg0ire:
Php might be configured differently depending on the distribution

jordisala1991:
After some research I found why is not restarted on PHP 5.3:

There is a constant that is only available on PHP 5.4 and forward: http://php.net/manual/es/reserved.constants.php#constant.php-binary

That constant is used to know when PHP needs to be restarted: https://github.com/composer/composer/blob/1.5/src/Composer/XdebugHandler.php#L117

Giving always `false` because PHP 5.3 does not have that constant.

It is documented on the composer test: https://github.com/composer/composer/blob/ff61ecd484d4e280e9892c13de3531a6a115aeed/tests/Composer/Test/XdebugHandlerTest.php#L21

But not in the official documentation (could be nice to know actually).

greg0ire:
Great job @jordisala1991 ! Should we downgrade Composer for 5.3 jobs ?
@sonata-project/contributors or should we simply drop 5.3 support? The amount of time wasted is getting ridiculous.

OskarStark:
👍  **drop** PHP 5.3 support
👎  **keep** PHP 5.3 support

core23:
Maybe just drop it for Travis or move it to allowed failures

greg0ire:
> Maybe just drop it for Travis or move it to allowed failures
Could generate a lot of support.

I think we should generalize this to "drop everything that is no longer maintained, meaning versions that do not even receive security patches"

jordisala1991:
> Great job @jordisala1991 ! Should we downgrade Composer for 5.3 jobs ?

not necessary. Everything else is working fine.

> @sonata-project/contributors or should we simply drop 5.3 support? The amount of time wasted is getting ridiculous.

Not sure if this is the right place to discuss it. I'm only trying to improve the build speed.

greg0ire:
> I'm only trying to improve the build speed.

Right, this is indeed not critical.

> Not sure if this is the right place to discuss it.

Indeed, the correct place is #232

Soullivaneuh:
This method disable xdebug globally, not only for composer.

BTW, PHP 5.3 will be dropped by the time with new majors, I don't think this change really worth it.

jordisala1991:
It disables xDebug until composer finishes install.

After that it is enabled again.

greg0ire:
> BTW, PHP 5.3 will be dropped by the time with new majors, I don't think this change really worth it.

On the stable branch, yeah... but what about the legacy branch?

Soullivaneuh:
> On the stable branch, yeah... but what about the legacy branch?

Still on it. But legacy branch should not get bunch of PRs.

Soullivaneuh:
Thanks a lot @jordisala1991

greg0ire:
If we force bugfixes to go through it, I'm afraid it will.

PR #281, opened 2017-09-06
Added docs folder to .gitattributes

core23:
We should remove all docs from the composer artifact.

### DIFF
--- a/project/.gitattributes.twig
+++ b/project/.gitattributes.twig
@@ -1,3 +1,4 @@
 .* export-ignore
 *.md export-ignore
 {{ tests_path }}/* export-ignore
+{{ docs_path }}/* export-ignore


greg0ire:
"git ignore" is a bit misleading :P

core23:
Ops 😊

Soullivaneuh:
Ignoring doc was already proposed loooong time ago by... myself. :wink: 

But this was refused. Please see: https://github.com/sonata-project/SonataAdminBundle/pull/3043

cc @rande

rande:
Documentation is almost important as code. So it should be kept in the composer artifact. If people does not have access to internet, they can still have access to the documentation.

greg0ire:
Documentation is probably not needed in production though.

rande:
@greg0ire this is not a valid argument, some classes are not required in production if the DI does not need them... but still they are in production. Documentation is not hurting anyone. Does not make you code bigger or unmaintainable. 

As a side note, sometime we get projets from production servers. We were more than happy when you get documentation too.

If you don't want files on production server you can also exclude `*.md` or `*.rst` while building your deployment artifact. Yes, composer should not be used in production.

greg0ire:
> Does not make you code bigger 

It makes the docker images a bit bigger, way bigger than the code itself if there are images

rande:
@greg0ire so the issue is not coming from composer, but from the way you build your docker image.

greg0ire:
Also, if every package does this, it means you get documentation even for things you don't really care about, like the dependencies of your dependencies. And personally, I never very rarely read docs from the vendors.

> the issue is not coming from composer, but from the way you build your docker image.

I'd really expect to be able to get a production ready version of my libs through Composer. People that need the txt version of the docs that badly have still an option: using `--prefer-source`.

> As a side note, sometime we get projets from production servers

:thinking: what ? :sweat_smile:

rande:
@greg0ire you don't code open source projects only for your needs (sorry!) and you cannot expect packages are production ready (what does that means btw ?).

We do something similar for cleaning up projects for production:

https://github.com/sonata-project/sandbox/blob/2.3/bin/qa_build_git.sh#L132-L146

You can add many rules you want to remove any files you don't want.

greg0ire:
> @greg0ire you don't code open source projects only for your needs (sorry!) and you cannot expect packages are production ready (what does that means btw ?).

I meant you don't have to filter anything out.

I understand your point of view, and I feel something is missing in the system like the possibility to download a docs archive for a given package. Not sure if it's an issue with Composer or with github.

rande:
@sonata-project/contributors however, I prefer to have a consensus on that point, so if every one want to remove the documentation, go for it. You might also want to remove test too.

Soullivaneuh:
> You might also want to remove test too.

It's already done AFAIK.

core23:
According to https://github.com/sonata-project/dev-kit/issues/314, we could remove the docs

Issue #278, opened 2017-08-26
Folder structure


core23:
Which folder structure should we follow? The current one or `src` and `test` / `tests`?

Also which name should we use `test` or`tests`?

**Pros for new structure**
- it allows to avoid cluttering the main autoloader
- clear differentiation between sources, docs and tests

**Cons**
 - changes the current structure

greg0ire:
I vote `src` and `tests`

core23:
Also which structure should we use for our docs? Should we keep them under (src)/Resources/docs or should we move them to the root inside a new `docs` folder?

core23:
@sonata-project/contributors Please vote for this. So we can prepare new major version with a new general folder structure.

core23:
👍 for  `src` and `tests`
👎 for `src` and `test`
🎉 for the current structure

core23:
👍 for a new `docs` folder
👎 for a new `doc` folder
🎉 for the current structure

Soullivaneuh:
Should be `tests` for running test files, and `test` for usable test classes.

Like it's done here: https://github.com/symfony/form

For the docs, it should be reviewed with @rande.

core23:
> Should be `tests` for running test files, and `test` for usable test classes.

That thing won't change. This thread is just for the folder structure for all runable test classes, not for test helpers / extensions of PHPUnit.

core23:
Please vote @jordisala1991 @Soullivaneuh @Th3Mouk

jordisala1991:
This structure applies to all bundles? Or only to libraries?

greg0ire:
I meant my vote for everything including bundles.

core23:
We should apply this to all bundles, so we can use the dev-kit more easily for all projects.

Soullivaneuh:
> That thing won't change. This thread is just for the folder structure for all runable test classes, not for test helpers / extensions of PHPUnit.

Not the issue. Symfony rules are prior to our owns on sonata-project. So this is not the question of a vote. `tests` should be used for testing files as `test` should be used for case test files.

Are your submitted PR according to this structure? If not, it must be changed.

Also this is maybe to late, but it would be better to have @rande advice about that. :neutral_face: 

Regards.

greg0ire:
Related: https://github.com/symfony/symfony-docs/issues/5691

Soullivaneuh:
@core23 Another question: Will this structure compatible for an eventual mono-repository integration of some of these projects?

[...]

PR #283, opened 2017-09-15
Added tests_path config key

core23:
The path for all test classes differs between the bundles, either it is `Tests` or `tests`, so we must have a config key for this.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -219,11 +219,13 @@ ecommerce:
       versions:
         symfony: ['2.8']
       docs_path: docs
+      tests_path: tests
     2.x:
       php: ['5.6', '7.0', '7.1']
       versions:
         symfony: ['2.8']
       docs_path: docs
+      tests_path: tests
 
 exporter:
   excluded_files:
@@ -236,6 +238,7 @@ exporter:
         doctrine_odm: ['1']
       services: [mongodb]
       docs_path: docs
+      tests_path: tests
     1.x:
       php: ['5.6', '7.0', '7.1']
       versions:
@@ -243,6 +246,7 @@ exporter:
         doctrine_odm: ['1']
       services: [mongodb]
       docs_path: docs
+      tests_path: tests
 
 formatter-bundle:
   branches:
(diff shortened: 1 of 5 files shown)


Soullivaneuh:
Would be better to decide which structure to use.

And it looks more like a psr-0/psr-4 usage difference than just a tests directory name, am I right?

core23:
Yes, it is the follow-up to https://github.com/sonata-project/dev-kit/issues/278. Do you have a better name for this key? @Soullivaneuh

Soullivaneuh:
Note: Waiting https://github.com/sonata-project/dev-kit/issues/278#issuecomment-330277074

core23:

src/Console/Command/AutoMergeCommand.php
```diff
@@ -120,7 +118,7 @@ private function mergeBranches(Package $package, array $projectConfig)
 
                     $this->io->warning($message);
                     $this->slackClient->attach([
-                        'text' => $message,
+                        'text'  => $message,
```
Soullivaneuh: These kind of change are wrong and are not related to this PR. Please revert.

core23: I removed these commits, but StyleCI won't let me merge this.

Also tried to manual trigger StyleCI, but it won't help :/

Soullivaneuh: This is weird. Array alignment should not be activated.

Are you up to date according to master?

core23: Did a fresh rebase for this.

Maybe StyleCI is confused, because we have a [repo config](https://github.com/sonata-project/dev-kit/blob/master/.php_cs) and the [project config](https://github.com/sonata-project/dev-kit/blob/master/project/.php_cs)?

Soullivaneuh: The repo config is just a symbolic link to the project config, so it should not be.

Soullivaneuh: I did a fresh analyse on master and StyleCI tells nothing: https://styleci.io/repos/55150410

Soullivaneuh:
Let's merge it to see what will be done.

PR #289, opened 2017-09-20
Improve merging strategy advices.

jlamur:
This is my understanding of how merging is currently done in the project.


Soullivaneuh:
Thank you @jlamur

Issue #285, opened 2017-09-20
Should we drop each vendor version support on minor releases?


Soullivaneuh:
Simple question, tiny issue.

:+1: To drop vendor support even for PHP and SF on minor release.
:-1: To keep as we do today. Drop version support on next releases only.

My thought:

:+1: Because it's semver compliant, and Doctrine maintainers already follow this way.
:+1: Because it would be **much** simpler to maintain
:-1: Because we don't have minor branches, so we can't provide bug-fixes for older minors supporting older vendor versions.

In any case, if we decide to do that, we must wait a bit after vendor release before dropping support. BTW, if we can support older version without a lot of pain, we should keep it to make transition smoother.

:warning: This does not affect this rules: We **DO NOT** drop version support of still maintained PHP and SF version and we **MUST** keep at least on LTS SF version support (currently 2.8) for concerned projects.

greg0ire:
To clarify, that would mean dropping php 5.3-5.5, and sf 2.3 on the stable branches, right? Will this policy apply on the current stable branches, or only starting with the next wave of major releases?

Soullivaneuh:
> To clarify, that would mean dropping php 5.3-5.5, and sf 2.3 on the stable branches, right?

Right.

> Will this policy apply on the current stable branches, or only starting with the next wave of major releases?'

I don't see why this can't be apply to current stable branches.

Soullivaneuh:
> BTW, if we can support older version without a lot of pain

Example: If PHP 7.2 goes out and our project perfectly work with PHP 7.0 and 7.1 without having to deal with BC condition (`method_exists`, `class_exists` etc...) we should not drop the support form composer right now.

greg0ire:
> I don't see why this can't be apply to current stable branches.

I'm all for it, just wanted to clarify

jordisala1991:
IMO if there is a good reason to drop some version of SF or PHP. We should be able to do it. Not as a: we will only be working with the latest version only, but to avoid things like having to deal with old PHP 5.3. Im +1 to have the power. But with a great power comes great responsibility

greg0ire:
I think we should only use it for versions that are no longer maintained.

Soullivaneuh:
@jordisala1991 Speciales rules for PHP and SF:

* PHP: Keep support for maintained versions (too late for 5.6, but we should not repeat this)
* Symfony: Support at least one LTS version (currently 2.8, nearly 3.4)

greg0ire:
[...]

PR #292, opened 2017-10-06
Stop testing unmaintained versions

greg0ire:
Refs #285

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -7,7 +7,7 @@ admin-bundle:
         sonata_core: ['3']
         sonata_block: ['3']
     3.x:
-      php: ['5.3', '5.4', '5.5', '5.6', '7.0', '7.1']
+      php: ['5.6', '7.0', '7.1']
       versions:
         symfony: ['2.3', '2.7', '2.8', '3.2', '3.3']
         sonata_core: ['3']
@@ -45,7 +45,7 @@ block-bundle:
         # https://travis-ci.org/sonata-project/SonataBlockBundle/jobs/131844587#L222
         #sonata_admin: ['3']
     3.x:
-      php: ['5.3', '5.4', '5.5', '5.6', '7.0', '7.1']
+      php: ['5.6', '7.0', '7.1']
       versions:
         symfony: ['2.3', '2.7', '2.8', '3.2', '3.3']
         sonata_core: ['3']
@@ -60,7 +60,7 @@ cache:
     master:
       php: ['7.1']
     1.x:
-      php: ['5.3', '5.4', '5.5', '5.6', '7.0', '7.1']
+      php: ['5.6', '7.0', '7.1']
 
 cache-bundle:
   branches:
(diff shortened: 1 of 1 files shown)


config/projects.yml
```diff
@@ -403,9 +405,9 @@ user-bundle:
         sonata_core: ['3']
         sonata_admin: ['3']
     3.x:
-      php: ['5.3', '5.4', '5.5', '5.6', '7.0', '7.1']
+      php: ['5.6', '7.0', '7.1']
       versions:
-        symfony: ['2.3', '2.7', '2.8']
+        symfony: ['2.8']
```
core23: Please add symfony 3.3, so we have at least one symfony 3 release remaining

config/projects.yml
```diff
@@ -387,9 +389,9 @@ translation-bundle:
         sonata_core: ['3']
         sonata_admin: ['3']
     2.x:
-      php: ['5.4', '5.5', '5.6', '7.0', '7.1']
+      php: ['5.6', '7.0', '7.1']
       versions:
-        symfony: ['2.3', '2.7', '2.8', '3.2', '3.3']
+        symfony: ['2.8', '3.1', '3.2']
```
core23: Symfony 3.1 is EOL

config/projects.yml
```diff
@@ -157,10 +157,10 @@ doctrine-mongodb-admin-bundle:
         symfony: ['2.8', '3.2', '3.3']
         sonata_admin: ['3']
     3.x:
-      php: ['5.3', '5.4', '5.5', '5.6']
+      php: ['5.6']
```
core23: What about PHP 7.1?

config/projects.yml
greg0ire: Woops that one was unintended

greg0ire: Had an old version, and made my changes before rebasing.

config/projects.yml
greg0ire: Is there a proof PR for that? I'm not sure the user bundle is supposed to work with sf3 in version 2

config/projects.yml
greg0ire: Give me a proof PR and I will add it.

config/projects.yml
core23: You're right. In that case 👍

Soullivaneuh:
@greg0ire I think we can drop PHP 5.6 and 7.0 too as doctrine does.

The only LTS constraint described in #285 is about Symfony.

Except maybe if another sonata dependent project does not work on php 7.1?

greg0ire:
On the stable branches? I think it's nice to keep at least the latest minor of php we started the branch with. I'm more ok with dropping 7.0 than 5.6... Although 7.0 still has active support, and 5.6 does not.

Soullivaneuh:
> On the stable branches? I think it's nice to keep at least the latest minor of php we started the branch with. I'm more ok with dropping 7.0 than 5.6... Although 7.0 still has active support, and 5.6 does not.

As discussed on slack, we keep support only on the green section: http://php.net/supported-versions.php

For 7.0, we will exceptionally drop it as it's already done on some projects.

config/projects.yml
```diff
@@ -22,7 +22,7 @@ admin-search-bundle:
         sonata_admin: ['3']
         ruflin_elastica: ['2']
     1.x:
-      php: ['5.6', '7.0']
+      php: ['7.0']
```
Soullivaneuh: 7.1 should be supported: https://github.com/sonata-project/SonataAdminSearchBundle/blob/9c3496449e4a6d21a8a196548b378bceb74fa9ad/composer.json#L20

PR #295, opened 2017-10-07
Add dependencies support rules

### LINKED ISSUE #285: Should we drop each vendor version support on minor releases?

Soullivaneuh:
From now, we can drop dependencies versions on minor releases.

These rules are here to clarify the constraints and **MUST** be the only
reference.

Closes #285 

Next: #292

### DIFF
--- a/project/CONTRIBUTING.md
+++ b/project/CONTRIBUTING.md
@@ -271,6 +271,18 @@ or postponed to next major release if BC is not possible.
 
 If you are not sure what should be done, don't hesitate to open an issue about your PR project.
 
+##### Dependency changes
+
+If you want to change some dependencies, here are the rules:
+
+- Don't add support for a version lower than the current one.
+- Don't change the highest supported version to a lower one.
+- Lower version dropping is accepted as a Backward Compatible change according to [semver][semver_dependencies_update],
+but some extra rules must be respected here:
+  - PHP versions that are under the [green zone][php_supported_versions] (actively maintained) **MUST NO** be dropped, even on master.
+  - If it's a Symfony package, at least the last LTS version **MUST** be supported, even on master.
+  - Generally, don't drop dependency version it it doesn't have a big impact on the code.
+
 ##### Legacy branches
 
 Legacy branches are **NOT** supported at all. Any submitted Pull Request will be immediately closed.
@@ -423,6 +435,17 @@ use the "Rebase and merge" feature.
 And finally, use your common sense : if you see a PR about a typo,
 or if there is a situation (faulty commit, revert needed) maybe you can merge it directly.
 
+#### Dependencies version dropping
+
+Do not merge any merge request dropping a dependency version support.
+To achieve that, mark them as `RTM`, and mention then on Slack when asking for a release.
+
+This rule should be applied for these reasons:
+
+- Even if it's semver compatible, we don't maintain minor branches.
+So it's preferable to give priority to bugfixes over version-dropping PRs.
+- Some dependencies need a dev-kit update. By the way, you can make a PR on dev-kit and link it to your own.
+
 ### Be nice to the contributor
 
 Thank them for contributing. Encourage them if you feel this is going to be long.
@@ -432,3 +455,5 @@ code yourself, or ping someone who can help.
 [sphinx_install]: http://www.sphinx-doc.org/en/stable/
 [pip_install]: https://pip.pypa.io/en/stable/installing/
 [sf_docs_standards]: https://symfony.com/doc/current/contributing/documentation/standards.html
+[semver_dependencies_update]: http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api
+[php_supported_versions]: http://php.net/supported-versions.php


project/CONTRIBUTING.md
```diff
@@ -271,6 +271,18 @@ or postponed to next major release if BC is not possible.
 
 If you are not sure what should be done, don't hesitate to open an issue about your PR project.
 
+##### Dependencies changes
+
+If you want to change some dependencies, here is the rules:
```
greg0ire: "here are"

project/CONTRIBUTING.md
```diff
@@ -271,6 +271,18 @@ or postponed to next major release if BC is not possible.
 
 If you are not sure what should be done, don't hesitate to open an issue about your PR project.
 
+##### Dependencies changes
```
greg0ire: "Dependency changes"

project/CONTRIBUTING.md
```diff
@@ -271,6 +271,18 @@ or postponed to next major release if BC is not possible.
...
+
+If you want to change some dependencies, here is the rules:
+
+- Don't add version support lower than the current one.
```
greg0ire: "Don't add support for a version lower than the current one"

project/CONTRIBUTING.md
```diff
@@ -271,6 +271,18 @@ or postponed to next major release if BC is not possible.
...
+- Don't change the highest version support to a lower one.
```
greg0ire: "the highest supported version"

project/CONTRIBUTING.md
Soullivaneuh: Well, you don't ever made changes about only one dependency, right?

project/CONTRIBUTING.md
```diff
@@ -271,6 +271,18 @@ or postponed to next major release if BC is not possible.
...
+- Lower version dropping is accepted as a Backward Compatible change according to [semver](semver_dependencies_update),
```
greg0ire: Shouldn't it be with square brackets everywhere? `[semver][semver_dependencies_update]` ?

Soullivaneuh: Indeed.

project/CONTRIBUTING.md
```diff
@@ -271,6 +271,18 @@ or postponed to next major release if BC is not possible.
...
+but some extra rules must be respected here:
+  - PHP versions that are under the [green zone](php_supported_versions) (actively maintained) **MUST NO** be dropped, even on master.
+  - If it's a Symfony package, at least the last LTS version **MUST** be supported, even on master.
+  - Generally, don't drop dependency version if this don't have a big impact on the code.
```
greg0ire: "don't drop a dependency version if it does not"

project/CONTRIBUTING.md
```diff
@@ -423,6 +435,17 @@ use the "Rebase and merge" feature.
...
+#### Dependencies version dropping
+
+Do not merge any merge request dropping a dependency version support.
+Add the `RTM` label and specify it while asking for a release on Slack.
+
+This rule should be applied for these reasons:
+
+- Even if it's semver compatible, we don't maintain minor branches.
+So it's preferable to release it at last time.
+- Some dependencies needs a dev-kit update. BTW, you can make a PR on dev-kit on link it to your own.
```
jlamur: `BTW` it's better to not use acronyms in this kind of document IMO.

project/CONTRIBUTING.md
```diff
@@ -423,6 +435,17 @@ use the "Rebase and merge" feature.
 And finally, use your common sense : if you see a PR about a typo,
 or if there is a situation (faulty commit, revert needed) maybe you can merge it directly.
 
+#### Dependencies version dropping
+
+Do not merge any merge request dropping a dependency version support.
+Add the `RTM` label and specify it while asking for a release on Slack.
```
greg0ire: what do you mean by "specify it" ?

project/CONTRIBUTING.md
greg0ire: You make several dependency changes

project/CONTRIBUTING.md
Soullivaneuh: Add links to the needed merge on the release section.

project/CONTRIBUTING.md
Soullivaneuh: Indeed.

project/CONTRIBUTING.md
```diff
@@ -423,6 +435,17 @@ use the "Rebase and merge" feature.
...
+#### Dependencies version dropping
+
+Do not merge any merge request dropping a dependency version support.
+Add the `RTM` (Ready To Merge) label and specify it while asking for a release on Slack.
+
+This rule should be applied for these reasons:
+
+- Even if it's semver compatible, we don't maintain minor branches.
+So it's preferable to release it at last time.
+- Some dependencies needs a dev-kit update. By the way, you can make a PR on dev-kit on link it to your own.
```
greg0ire: "need"

greg0ire: "or link it"?

Soullivaneuh: and

project/CONTRIBUTING.md
```diff
@@ -432,3 +455,5 @@ code yourself, or ping someone who can help.
 [sphinx_install]: http://www.sphinx-doc.org/en/stable/
 [pip_install]: https://pip.pypa.io/en/stable/installing/
 [sf_docs_standards]: https://symfony.com/doc/current/contributing/documentation/standards.html
+[semver_dependencies_update](http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api)
```
greg0ire: Wrong syntax I think

project/CONTRIBUTING.md
```diff
@@ -271,6 +271,18 @@ or postponed to next major release if BC is not possible.
...
+
+If you want to change some dependencies, here are the rules:
+
+- Don't add support for a version lower than the current one.
+- Don't change the highest supported version to a lower one.
+- Lower version dropping is accepted as a Backward Compatible change according to [semver][semver_dependencies_update],
+but some extra rules must be respected here:
+  - PHP versions that are under the [green zone][php_supported_versions] (actively maintained) **MUST NO** be dropped, even on master.
+  - If it's a Symfony package, at least the last LTS version **MUST** be supported, even on master.
+  - Generally, don't drop dependency version if this don't have a big impact on the code.
```
greg0ire: "it it doesn't have a big impact"

Soullivaneuh:
@greg0ire Is it ok for you?

greg0ire:
:shipit:

PR #299, opened 2017-10-08
Switch from precise to trusty

greg0ire:
Since we dropped 5.3, we no longer need to stick with precise.


Soullivaneuh:
Ref: https://github.com/sonata-project/dev-kit/commit/4016f71d971a150c655ecef0599ece897136be72

PR #300, opened 2017-10-09
PHPCR branches update

Soullivaneuh:


dbu:
i guess for master we can drop everything but symfony 3.4 / 4.0 once that is released. but thats no hurry, can't do it before that is released.

Soullivaneuh:
Yes, this can be updated later.

@dbu BTW, did you plan to update the changelog format accordingly to the contributing document?

PR #302, opened 2017-10-09
Improve ISSUE_TEMPLATE.

jlamur:
Some people tend to not put the result of their commands
inside the code block.

### DIFF
--- a/project/.github/ISSUE_TEMPLATE.md
+++ b/project/.github/ISSUE_TEMPLATE.md
@@ -14,12 +14,14 @@
 
 ```
 $ composer show --latest 'sonata-project/*'
+# Put the result here.
 ```
 
 #### Symfony packages
 
 ```
 $ composer show --latest 'symfony/*'
+# Put the result here.
 ```
 
 #### PHP version


project/.github/ISSUE_TEMPLATE.md
```diff
@@ -14,12 +14,14 @@
 
 ```
 $ composer show --latest 'sonata-project/*'
+# Put the result here.
```
core23: `<!-- Put the result here. -->`

Soullivaneuh: Not sure it will work with the code block.

core23: The `#` infront should mark the following text as a comment not a headline

jlamur: Only if you consider this code block as a markdown code block.

The ambiguity can be resolved by setting the syntax to `bash`, like this:
````
```bash
$ composer show --latest 'sonata-project/*'
# Put the result here.
```
````

This renders as:
```bash
$ composer show --latest 'sonata-project/*'
# Put the result here.
```

core23: Just discovered, that this is inside the bash code snippet.

Maybe modify it to something like this, because the result is not part of the bash script

 ````
```bash
 $ composer show --latest 'symfony/*'
 ```
<!-- Put the result here. -->
````

greg0ire: This is not a bash script, `$` has no meaning in bash. This is a shell with a prompt and an output, keep them together please.

Soullivaneuh: @core23 Your example is confusing. You will have result put outside the block.

Soullivaneuh: IMHO: The "put the result here" is not necessary, it's quite obvious, isn't it?

If you really want this, I see two solutions:

* Keep the `#` comment as is
* Do not put anything here but something like `<!-- Run the following command and place the result on the correponding blocks -->` just after `### Environment` section.

core23: I'm confuse too 🤣  Don't know what I was thinking about. Just forget that I was here

PR #305, opened 2017-10-14
Add composer flags to every update command, remove quiet flag

jordisala1991:
I made two changes on the first composer update command.

1. Remove quiet flag: now that we are always doing 2 composer update on every build the first one get all the dependencies installes (most of the time with the correct versions), the second one only fixes the problems with --prefer-lowest and --platform-reqs bugs.

2. Add composer_flags to the first command. If this is not present, this build does not pass. Note that it does make sense to install --prefer-lowest both times:

https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/230

### DIFF
--- a/project/.travis/install_test.sh
+++ b/project/.travis/install_test.sh
@@ -19,6 +19,6 @@ chmod u+x "${HOME}/bin/coveralls"
 # To be removed when these issues are resolved:
 # https://github.com/composer/composer/issues/5355
 # https://github.com/composer/composer/issues/5030
-composer update --prefer-dist --no-interaction --prefer-stable --quiet --ignore-platform-reqs
+composer update --prefer-dist --no-interaction --prefer-stable --ignore-platform-reqs ${COMPOSER_FLAGS}
 
 composer update --prefer-dist --no-interaction --prefer-stable ${COMPOSER_FLAGS}


project/.travis/install_test.sh
```diff
@@ -19,6 +19,6 @@ chmod u+x "${HOME}/bin/coveralls"
 # To be removed when these issues are resolved:
 # https://github.com/composer/composer/issues/5355
 # https://github.com/composer/composer/issues/5030
-composer update --prefer-dist --no-interaction --prefer-stable --quiet --ignore-platform-reqs
+composer update --prefer-dist --no-interaction --prefer-stable --ignore-platform-reqs ${COMPOSER_FLAGS}
```
Soullivaneuh: See this blame:

This was a hack and should not have the same behavior than the other one.

BTW this change was an error IMHO: https://github.com/sonata-project/dev-kit/commit/3e98f514db3ec02852ea1dd0f0ab5f19e0091eaa#diff-ce628c33291abbf4b03aaa72535c90f2L20

Soullivaneuh:
@jordisala1991 Please avoid yo make test on dev-kit PR. Those PR are automatically merged when they are green. This is dangerous.

project/.travis/install_test.sh
jordisala1991: I don't know any other way to make it work with mongoDBAdminBundle.

Soullivaneuh: I'm talking about the if condition. BTW I update the broken blame link.

jordisala1991: Both platform-requirements and prefer-lowest have some bugs on composer and require a composer.lock file to make them work

Soullivaneuh: @jordisala1991 Yes. This is why this composer was runned only whe --prefer-lowest is set.

But the condition was removed. Why?

jordisala1991: There are 2 things: The old code won't work also on this case and it was removed because when you use php7 you need a package that provides compatibility with `ext-mongo` (because it was removed on php7).

This other composer bug: https://github.com/composer/composer/issues/5030

Soullivaneuh: Two things:

- It would be great to add a condition even if it will be note the same (Maybe a custom ENV?) because not all sonata project needs this.
- This modifcation is not good because you will run a --prefer-lowest twice and because of the first composer issue, we have to run a classic composer update first to get the `.lock` file.

jordisala1991: I see, this change is not good at all, but not sure why sonata-project/SonataDoctrineMongoDBAdminBundle#230 only pass if you do double composer update with --prefer-lowest

jordisala1991: Should we do a variable on projects.yml or ENV var on travis to see which behavior to choose?

1. only double composer update on --prefer-lowest
2. do always double composer update for the --ignore-platform-reqs flag

Soullivaneuh: @jordisala1991 I don't see how double `prefer-lowest` will fix the issue, it looks insane. :D

Did you try with `composer update` and then `composer update --prefer-lowest`?

jordisala1991: yes, I have tried that on SonataDoctrineMongoDBAdminBundle, currently with the changes provided by SonataCI it does not work (see sonata-project/SonataDoctrineMongoDBAdminBundle#230 PR). If you try adding a double `--prefer-lowest` it starts working again :)

Soullivaneuh: The `composer update` and `composer update --prefer-lowest` is already a big workaround. Would be better to find why it's doing this and fix it.

BTW, I'll make another PR to restore this: https://github.com/sonata-project/dev-kit/commit/3e98f514db3ec02852ea1dd0f0ab5f19e0091eaa#diff-ce628c33291abbf4b03aaa72535c90f2L20

Soullivaneuh: @jordisala1991 BTW, don't the version support reduction solve the issue?

jordisala1991: I tried everything and not.

I think this PR should be closed because this is not the correct solution, but we need to find one, because restoring that commit will not make the build pass (it could make other builds fail) (even though it is probably a good idea).

PR #312, opened 2017-10-24
 Deprecation rules improvement

Soullivaneuh:
- The trigger_error usage is not necessary while using `@deprecated` tag.
  Recent versions of Symfony properly handle the tag.
- The `NEXT_MAJOR` keyword should not be used on deprecation message.
  A search must be done on `@deprecated`, `E_USER_DEPRECATED` and `NEXT_MAJOR` before new major.
- Deprecation can and must be done on service definiton if concerned.
  This is now possible since Symfony 2.8.

Also the documentation is now more accurate about rules, especially for the upgrade note.

### DIFF
--- a/project/CONTRIBUTING.md
+++ b/project/CONTRIBUTING.md
@@ -263,8 +263,9 @@ public function baz()
 }
 ```
 
-In that case, unit tests might show your deprecation notice. You must mark such tests with the `@group legacy` annotation,
-and if need be, isolate them in a new test method that can simply be removed in the non-BC PR.
+In that case, unit tests might show your deprecation notice.
+You must mark such tests with the `@group legacy` annotation and if need be,
+isolate them in a new test method that can simply be removed in the non-BC PR.
 
 Be aware that pull requests with BC breaks could be rejected
 or postponed to next major release if BC is not possible.
--- a/project/CONTRIBUTING.md
+++ b/project/CONTRIBUTING.md
@@ -232,8 +232,62 @@ interface BarInterface
 }
 ```
 
-In some cases, you will have the possibility to warn the user that things will change,
-and recommend a new way of doing things. You can do so by triggering the dedicated kind of error, like this:
+##### Deprecation
+
+In some cases, you might have some code parts that might become superseded by others, but could still be used by the end user.
+If the concerned code is not tagged as `internal`, it must be deprecated on the stable branch, then removed.
+
+If an alternate usage solution is possible, you **MUST** provide it in the deprecation message.
+
+The deprecated minor version **MUST NOT** be provided. Use `x` instead. It will be updated when releasing.
+
+Any deprecation **MUST** be documented in the corresponding `UPGRADE-[0-9].x.md`.
+The documentation **MUST** be filled inside the top **unreleased** section with a sub title.
+
+The `NEXT_MAJOR` tag should not be used for deprecation.
+The `@deprecated` and `E_USER_DEPRECATED` key will be searched for before releasing the next major version.
+
+You have three ways to deprecate things.
+
+For class definitions, methods (or first level functions) and properties, use the `@deprecated` tag: 
+
+```php
+/**
+ * @deprecated since 42.x, to be removed in 43.0. Use Shiny\New\ClassOfTheMonth instead.
+ */
+final class IAmOldAndUseless
+{
+}
+
+final class StillUsedClass
+{
+    /**
+     * @deprecated since 42.x, to be removed in 43.0.
+     */
+    public $butNotThisProperty;
+
+    /**
+     * @deprecated since 42.x, to be removed in 43.0.
+     */
+    public function iAmBatman()
+    {
+        echo "But this is not Gotham here.";
+    }
+}
+```
+
+If the deprecated thing is a service, you **MUST** specify it on the service definition:
+
+```xml
+<service id="sonata.block.old" class="Sonata\Block\Old">
+    <argument type="service" id="security.token_storage" />
+    <deprecated>The "%service_id%" service is deprecated since 42.x and will be removed in 43.0.</deprecated>
+ </service>
+ ```
+
+More info: http://symfony.com/blog/new-in-symfony-2-8-deprecated-service-definitions
+
+For everything else, not managed by the `@deprecated` tag, you **MUST** trigger a deprecation message.
 
 ```php
 <?php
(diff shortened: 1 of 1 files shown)


project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
 }
 ```
 
-In some cases, you will have the possibility to warn the user that things will change,
-and recommend a new way of doing things. You can do so by triggering the dedicated kind of error, like this:
+##### Deprecation
+
+In some cases, you might have some code part being useless.
```
greg0ire: What do you mean? Why shouldn't we remove it directly if it is useless. To me, useless = cannot be reached.

project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
...
+If the concerned code is not tagged as `internal`, it must be deprecated on stable branch, then removed.
+
+If an alternate usage solution is possible, you **MUST** provide it on the deprecation message.
```
greg0ire: "in the"

project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
...
+
+The deprecated minor version **MUST NOT** be provided. Use `x` instead. It will be updated on releasing.
```
greg0ire: I think "when releasing" might be better, I'm not 100% sure "on releasing" is correct.

project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
...
+
+Any deprecation **MUST** be documented on the correponding `UPGRADE-[0-9].x.md`.
+The documentation **MUST** be filled on the top **not** released section with a sub title.
```
greg0ire: unreleased

project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
...
+
+The `NEXT_MAJOR` tag should not be used for deprecation.
+The `@deprecated` and `E_USER_DEPRECATED` key will be matched for next major release.
```
greg0ire: "matched for next major release" => "searched for before releasing the next major version"?

project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
...
+     * @deprecated since 42.x, to be removed in 43.0.
+     */
+    public function iAmBatman()
+    {
+        echo "But this is not Gotham here.";
+    }
+}
+```
+
+If the deprecated thing is a service, you **MUST** specify it on the service definition:
```
greg0ire: Oh right we can start doing that now! :+1:

project/CONTRIBUTING.md
Soullivaneuh: Example: When we moved `Admin` to `AbstractAdmin`, the `Admin` class is now useless. But can be still used by the end user.

Did you got my point? How would you explain it differently?

project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
...
+```xml
+<service id="sonata.block.old" class="Sonata\Block\Old">
+    <argument type="service" id="security.token_storage" />
+    <deprecated>The "%service_id%" service is deprecated since 42.x and will be removed in 43.0.</deprecated>
+ </service>
+ ```
+
+More info: http://symfony.com/blog/new-in-symfony-2-8-deprecated-service-definitions
+
+For the rest, not managed by the `@deprecated` tag, you **MUST** trigger a deprecation message.
```
greg0ire: "the rest" => "everything else"

project/CONTRIBUTING.md
Soullivaneuh: I'll trust you on this one. :wink:

Soullivaneuh:
@greg0ire all fixed.

project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
 }
 ```
 
-In some cases, you will have the possibility to warn the user that things will change,
-and recommend a new way of doing things. You can do so by triggering the dedicated kind of error, like this:
+##### Deprecation
+
+In some cases, you might have some code parts might become superseded by others, but could still be used by the end user.
```
greg0ire: Sorry I forgot a word: "that might become"

project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
...
+If the concerned code is not tagged as `internal`, it must be deprecated on stable branch, then removed.
```
greg0ire: "the stable branch"

project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
...
+
+If an alternate usage solution is possible, you **MUST** provide it in the deprecation message.
+
+The deprecated minor version **MUST NOT** be provided. Use `x` instead. It will be updated when releasing.
+
+Any deprecation **MUST** be documented on the correponding `UPGRADE-[0-9].x.md`.
```
greg0ire: "in the corre**s**ponding"

project/CONTRIBUTING.md
```diff
@@ -232,8 +232,62 @@ interface BarInterface
...
+The documentation **MUST** be filled on the top **unreleased** section with a sub title.
```
greg0ire: on => inside?

project/CONTRIBUTING.md
```diff
@@ -248,26 +302,14 @@ if (/* some condition showing the user is using the legacy way */) {
...
- */
-public function baz()
-{
-}
-```
+Note that the `trigger_error` usage is not necessary if the `@deprecated` tag is used.
 
-In that case, unit tests might show your deprecation notice. You must mark such tests with the `@group legacy` annotation,
-and if need be, isolate them in a new test method that can simply be removed in the non-BC PR.
+On deprecation case, unit tests might show your deprecation notice.
```
greg0ire: "in the case of a deprecation"

stof:
> The trigger_error usage is not necessary while using @deprecated tag.
> Recent versions of Symfony properly handle the tag.

that's not entirely true. Symfony will warn automatically when you extend a deprecated class or implement a deprecated interface. But it will not warn when *using* a deprecated class.

Soullivaneuh:
@stof Oh, right, so bad.

Just by curiosity, is it planned to be handled when using a deprecated class one time?

I'll update the documentation.

stof:
@Soullivaneuh short answer is: no.

I can provide a longer answer explaining why later if you want (but I don't have time to write the full explanation right now)

Soullivaneuh:
I'm interested to the long answer but do what you can! :wink:

PR #313, opened 2017-10-24
Do the composer lowest trick only when lowest is requested

Soullivaneuh:
Introduced in #238, It's currently done on each Travis job of each repo, this is not the wanted behavior.


jordisala1991:
I think it is a good idea, but probably it will make some builds fail (the ones that use mongodb)

Soullivaneuh:
The mongodb case is already broken as you tell and must be fixed differently.

PR #320, opened 2017-10-30
Add .md files to archives.

jlamur:
Composer archives did not have README.md, UPGRADE-*.md, ...
files.

### DIFF
--- a/project/.gitattributes.twig
+++ b/project/.gitattributes.twig
@@ -1,3 +1,2 @@
 .* export-ignore
-*.md export-ignore
 Tests/* export-ignore


Soullivaneuh:
It should be voted before merge.

jlamur:
If you keep the documentation, why would you not keep, at least, the CHANGELOG and UPGRADE files?

Soullivaneuh:
For my opinion, the only reference should be on the GitHub repo with the last update.

greg0ire:
> In my opinion, the only reference should be on the GitHub repo with the last update.

What do you mean by that?

I discussed with @rande last week and am no longer sure about this "docs exclusion" thing, that's why I didn't vote yet.

Soullivaneuh:
> What do you mean by that?

I mean the `.md` files should not be read form the local installation but from GitHub to be sure to have the last do update.

But this is only my opinion, and my vote. :wink:

jlamur:
Closing because we removed the docs from archives. It seems consistent to not have `*.md` files in archives too.

PR #326, opened 2017-11-04
Fix StyleCI configuration

Soullivaneuh:


Soullivaneuh:
Oh I got this one @core23. Didn't see you move this to a twig file. We can't do that.

core23:
That's the project config. 

Maybe we should keep two `.styleci.yml` files? One for the dev-kit and one for the projects?

Soullivaneuh:
Let's stick on both `tests` and `Tests` folder exclusion. It wont hurt anybody,

PR #330, opened 2017-11-07
Leave the Hacktoberfest

core23:
The October is over


Soullivaneuh:
We might find a better way than editing the code every year...

PR #333, opened 2017-11-09
Drop hhvm

greg0ire:
It adds complexity, and this project is being dropped by everyone
because it does not seem to be going in a good direction.


Soullivaneuh:
Will be merged later as the current projects "works" without this change (it's on allowed failure)

PR #335, opened 2017-11-10
Revert "Use the weak vendors mode"

core23:
This mode is risky and will fail some build, because there is a bug that will be fixed in symfony 3.4.

Refs: https://github.com/sonata-project/SonataPageBundle/pull/895

### DIFF
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -34,7 +34,7 @@ cache:
 env:
   global:
     - PATH="$HOME/.local/bin:$PATH"
-    - SYMFONY_DEPRECATIONS_HELPER=weak_vendors
+    - SYMFONY_DEPRECATIONS_HELPER=weak
     - TARGET=test
     - UPSTREAM_URL=https://github.com/sonata-project/{{ repository_name }}.git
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
> because there is a bug that will be fixed in symfony 3.4

Do you have the PR link of the fix?

@greg0ire Can't we just use phpunit-bridge `^3.4`? I don't think we need to follow Symfony requirement for this part.

core23:
This might fix it: symfony/symfony#24864

core23:
phpunit-bridge 3.4 doesn't exist atm

Soullivaneuh:
> This might

Might?

> phpunit-bridge 3.4 doesn't exist atm

Will be a 3.x fix: https://github.com/symfony/symfony/pull/24864#issuecomment-343504926

greg0ire:
rebased on 3.3

greg0ire:
I'm not confident it fixes this issue, and I'm not confident it will be merged quickly, so reverting sounds like a good idea. I'll make sure to make proof PRs if I want to add it back.

Soullivaneuh:
Okay let's do it then.

PR #341, opened 2017-11-15
Add travis wait

jordisala1991:
What do you think on having travis_wait for the `install_{TARGET}`?

We need atleast 30 minutes as I tested here: https://github.com/sonata-project/ecommerce/pull/480

Bad things about having this one:

- We don't have any logs of that bash script until it is finished executing, because of the travis_wait.
- most of the bundles don't need it, there are just a few ones and only on the `--prefer-lowest` build

### DIFF
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -80,7 +80,7 @@ before_install:
   - if [ -x .travis/before_install_${TARGET}.sh ]; then .travis/before_install_${TARGET}.sh; fi;
 
 install:
-  - if [ -x .travis/install_${TARGET}.sh ]; then .travis/install_${TARGET}.sh; fi;
+  - if [ -x .travis/install_${TARGET}.sh ]; then travis_wait 30 .travis/install_${TARGET}.sh; fi;
 
 before_script:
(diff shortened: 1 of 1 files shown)


greg0ire:
@jordisala1991 I think we should fix the root cause.

greg0ire:
Also, people are going to get crazy if they have to wait 30 minutes to get a build status :P

jordisala1991:
> Also, people are going to get crazy if they have to wait 30 minutes to get a build status :P

There are a few bundles that need this, and I agree it is a pain to wait for 30 minutes. But at the moment, those bundles are unable to do a single build, so if we don't get to the root cause, I prefer to wait 30 minutes rather than not be able to build at all.

greg0ire:
I'm looking into it right now, starting with the comment bundle. The issue might be with easy-extends-bundle.

greg0ire:
I think making a new release and requiring might help: https://github.com/sonata-project/SonataEasyExtendsBundle/compare/2.2.0...2.x

jordisala1991:
Closing it because it should be fixed releasing other bundles

PR #342, opened 2017-11-15
Reverse config for bundles without the new structure

jordisala1991:
It is easier to handle the 4 bundles that need a PR to change its folder structure.

This change is according to this: https://github.com/orgs/sonata-project/projects/3


Soullivaneuh:
Thanks!

PR #344, opened 2017-11-18
Test Symfony3 on bundles that are already using it

jordisala1991:
SonataCommentBundle, SonataNewsBundle, SonataPageBundle
are using SF3 on composer.json but it is not on travis builds

Proof on composer.json:

https://github.com/sonata-project/SonataPageBundle/blob/3.x/composer.json
https://github.com/sonata-project/SonataCommentBundle/blob/3.x/composer.json
https://github.com/sonata-project/SonataNewsBundle/blob/3.x/composer.json

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -102,14 +102,14 @@ comment-bundle:
     master:
       php: ['7.1']
       versions:
-        symfony: ['2.8']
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
Looks good, thanks!

PR #350, opened 2017-11-22
Add .x-dev to package specific dev-master builds

jordisala1991:
With this we can get more greens on the allowed failures builds. Why?

Because on some cases, those builds require a version upper than just `3` of Corebundle for example: 
https://travis-ci.org/sonata-project/SonataTranslationBundle/builds/303615904

With this change we are aliasing with the `3.x-dev` version which is the latest and cannot fail due to a requirement on composer.json

Proof:
https://travis-ci.org/sonata-project/SonataTranslationBundle/builds/306030700
https://github.com/sonata-project/SonataTranslationBundle/pull/205

*Note*: Do not look at `--prefer-lowest` build, it is unrelated.

### DIFF
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -59,7 +59,7 @@ matrix:
 {% endfor %}
 {% if package_versions|length > 0 %}
     - php: '{{ target_php|default(php|last) }}'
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
Thank you.

PR #355, opened 2017-12-04
Allow multiline signatures

### LINKED ISSUE #311: About multi-line method signature
Before all, I have to notice @sonata-project/contributors: I'm mentally fine, and @greg0ire didn't put a :gun: to my head to make this issue, don't worry. :trollface: 

But [his article](https://dev.to/greg0ire/the-case-for-multiline-signatures-dl6) is interesting, and since all our project are now under PHP 7+ and because he annoyed me every day with this rule :trollface: :heart:, I want to clarify it and make a consensus.

Also, I always told we must follow Symfony and PSR coding standard and don't loose time to have your own coding style except for important things, and **this is still applied**.

This case seems to be very important and I want to have a closing debate with this issue as reference.

So, since PHP 7 and advanced type hinting is here, it would be interesting to re-consider multi-line method signature to avoid the very long lines on some PHP code. I let you read the article for the details.

I would ask a lot of votes of @sonata-project/contributors about this issue to validate it, or not.

Should we go to multi-lines signature? Before voting, please not that if we want to do this:

1. All the method **MUST** be multi-lined. Or at least, starting with more than one parameter.
2. We **MUST** have a working fixer rule for that. This means, if the rules is available under PHP-CS-Fixer v2, we **MUST** get rid of StyleCI and find a better way.
3. Concerning the first two rules, we **MUST NOT** apply the multi-line rule before having this conditions filled.
4. No rule change for PHPDoc. It **MUST** be fully completed for now. This is a place for *maybe* another debate.

So, now the vote:

:+1: To have multi-lines method signature
:-1: To keep the Symfony rules as we are doing now
:confused: if you don't care at all.

Personal opinion: I think we may go to multi-line method signature, event this is not a big deal for me. Doctrine also does it AFAIK.

Regards.

Note: I will wait also @rande approval about that.

### LINKED ISSUE #349: Multiline signature counter proposal
@Soullivaneuh doesn't like the fact that I hijack his issue, so here's a new one, might be clearer / better anyway indeed.

So here is my counter-proposal: allow multiline signatures when the line is longer than 120 characters and do not wait for a fixer to exist before allowing it. That's sane and simple. Please vote for or against.

My main point is that you should not confuse consistency, and deliberately not using one tool of your toolbox. You don't always have to choose, just like for `sprintf` vs encapsed strings vs concatenation, some are better suited to some situations.

greg0ire:
Closes #311, Closes #349

### DIFF
--- a/project/CONTRIBUTING.md
+++ b/project/CONTRIBUTING.md
@@ -87,6 +87,8 @@ protected function bar($foo, string $name, int $limit)
 }
 ```
 
+Please also note that multiline signatures are allowed when the line is longer than 120 characters.
+
 #### The documentation
 
 The documentation is mostly written with the `rst` format, and can be found in the `{{ docs_path }}` directory.


project/CONTRIBUTING.md
```diff
@@ -87,6 +87,8 @@ protected function bar($foo, string $name, int $limit)
 }
 ```
 
+Please also note that multiline signatures are allowed when the line is longer than 120 characters.
```
Soullivaneuh: @greg0ire AFAIK, you are working on a fixer for that. Maybe could you open an issue here to remove this line when the fixer will be merged and integrated?

greg0ire: @Soullivaneuh done

PR #357, opened 2017-12-04
User bundle v4

Soullivaneuh:


Soullivaneuh:
Need #358 to get merged an deployed first.

PR #358, opened 2017-12-04
Get rid of StyleCI

Soullivaneuh:


project/Makefile
```diff
@@ -13,6 +13,7 @@ lint:
 	find . \( -name '*.xml' -or -name '*.xliff' \) \
 		-not -path './vendor/*' -not -path './Resources/public/vendor/*' -type f \
 		-exec xmllint --encode UTF-8 --output '{}' --format '{}' \;
+	php php-cs-fixer.phar fix --verbose
```
greg0ire: Consider adding

```
php-cs-fixer.phar:
	wget https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/download/v2/php-cs-fixer.phar

lint: php-cs-fixer.phar
```
so that it works locally

Soullivaneuh: I won't implement installation on Makefile, it was not done for phpunit. I'll made global call instead. Installation should not rely to us.

Soullivaneuh:
StyleCI must be deactivated **before** getting this PR merged.

project/Makefile
greg0ire: A global call sounds good!

PR #363, opened 2017-12-06
Remove php70 rule from general rules

jordisala1991:
This rule should be part of `PHP70Migration` since it is only allowed to use it under php 7.0 as minimum

### DIFF
--- a/project/.php_cs.dist
+++ b/project/.php_cs.dist
@@ -33,7 +33,6 @@ $rules = [
     'ordered_class_elements' => true,
     'ordered_imports' => true,
     'phpdoc_order' => true,
-    'ternary_to_null_coalescing' => true,
     // To be tested before insertion:
 //    'strict_comparison' => true,
 //    'strict_param' => true,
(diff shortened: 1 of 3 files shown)


project/.php_cs.dist
```diff
@@ -49,12 +48,14 @@ if (\PHP_VERSION_ID >= 50600) {
...
 if (\PHP_VERSION_ID >= 70000) {
     $rules = array_merge($rules, [
+        '@PHP70Migration' => true,
         '@PHP70Migration:risky' => true,
         '@PHPUnit60Migration:risky' => true,
     ]);
 }
 if (\PHP_VERSION_ID >= 70100) {
     $rules = array_merge($rules, [
+        '@PHP71Migration' => true,
```
jordisala1991: This one is causing travis to fail because `const` does not have visibility defined. Do I change dev-kit code also or remove this line?

jordisala1991: Done

PR #366, opened 2017-12-06
Dynamic phpcs

greg0ire:
This avoids relying on the current php version, that might very well differ from the target php version


greg0ire:
Dang, that was a symlink?!

Soullivaneuh:
@greg0ire Yes it was, and I would like to keep only one ruleset.

Plus, your solution is not exportable at all if one day we want to move those configuration to something like `sonata-project/coding-standards`. Even if yes, this is an another subject.

What is the issue with the current configuration?

jordisala1991:
The issue with the current configuration is:

Someone using php 7.1 wants to do a PR on current stable (php 5.6), it does his works, and creates the pull request. The linter fails because he misses some of our coding standards. We tell him to apply php-cs-fixer. He does, bum! all builds fails because he "misses" to remove 2 ifs statements from the `.php-cs.dist` file.

project/.php_cs.dist.twig
```diff
@@ -33,33 +33,25 @@ $rules = [
...
-if (\PHP_VERSION_ID >= 70100) {
-    $rules = array_merge($rules, [
-        '@PHP71Migration' => true,
-        '@PHP71Migration:risky' => true,
-        'compact_nullable_typehint' => true,
-    ]);
-}
 
 $finder = PhpCsFixer\Finder::create()
     ->in(preg_match('/dev-kit\/project$/', __DIR__, $matches) ? __DIR__.'/..' : __DIR__)
```
jordisala1991: This trick is no longer needed if we do it this way. 

Same for the dev-kit configuration

greg0ire:
> Plus, your solution is not exportable at all if one day we want to move those configuration to something like sonata-project/coding-standards. Even if yes, this is an another subject.


Not very important at the moment, the big issue I want to address here is to make it easy for anyone to fix style issues. Right now it's a pain, for the reason @jordisala1991 explained.

Soullivaneuh:
I understand. Still, the duplicate configuration should be avoided.

greg0ire:
> I understand. Still, the duplicate configuration should be avoided.

Can we merge this now and you fix it later if you want to?

Soullivaneuh:
> Can we merge this now and you fix it later if you want to?

This will just be forgotten, and the configuration will later be forgotten.

Can't we make dynamic on PHP side? With a `.env` file on each repository for example?

greg0ire:
> Can't we make dynamic on PHP side? With a .env file on each repository for example?

I tried a trick with env vars in the makefile, but it looks fragile. See https://github.com/sonata-project/SonataAdminBundle/pull/4829#discussion_r155348463 (it's an outdated discussion now)

jordisala1991:
If you have a .env in each branch pointing out to `{{ php|first }}` and you load somehow that env in .php-cs.dist It could work. It seems to me like a lot to make dev-kit work without copying config.

greg0ire:
> It seems to me like a lot to make dev-kit work without copying config.

To me too, and that's key: dev-kit is here to help, its influence should be invisible apart from comments that tell ppl not to edit files directly. It should not leak any constraint into Sonata.

jordisala1991:
If the problem is getting dev-kit cs fixer outdated because it could be forgotten. Let's make this rule:

If you want to add something to project cs-fixer, first you need to get it working on dev-kit

Soullivaneuh:
> If you want to add something to project cs-fixer, first you need to get it working on dev-kit

This rule is implicit and already exists. Making those file separate is the best way to forget to do it.

But as you want, let's keep this.

@greg0ire Because GH is telling me to f**k myself:

image

PR #368, opened 2017-12-06
Always have a lint build

jordisala1991:
It makes no sense to remove lint build if there is no need to do a docs build.

### DIFF
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -51,10 +51,8 @@ matrix:
     - php: '{{ php|last }}'
       env: TARGET=docs
 {% endif %}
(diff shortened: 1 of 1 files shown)


project/.travis.yml.twig
```diff
@@ -51,10 +51,8 @@ matrix:
     - php: '{{ php|last }}'
       env: TARGET=docs
 {% endif %}
-{% if docs_target %}
```
greg0ire: Uuuuh yeah that looks funny indeed

Soullivaneuh: :wine_glass: :beers:

PR #369, opened 2017-12-07
Fix install for mongodb

jordisala1991:
This finally fixes the issue in a cleaner way. We don't need anymore the `--ignore-platform-reqs`, which is good.

And we only modify things for the builds that need mongodb.

Proof: https://travis-ci.org/sonata-project/SonataDoctrineMongoDBAdminBundle/jobs/312910083

And the idea comes from here:

https://github.com/alcaeus/mongo-php-adapter/pull/207

Thanks @alcaeus


alcaeus:
Happy to be of service - with the release of 7.2 you don't want to use `--ignore-platform-reqs` on PHP 7.1 as it could lead to invalid packages being installed.

PR #371, opened 2017-12-09
Makefile improvements

greg0ire:


greg0ire:
Proof PR: https://github.com/sonata-project/SonataAdminBundle/pull/4829

jordisala1991:
Can you explain how we use this makefile? I have no idea what .PHONY is.

Not sure if we should have some docs related on the usage of this makefile for the contributors

greg0ire:
. PHONY means this is not a file, don't bother checking the modification date

jordisala1991:
So the commands are:

```
Those are only for check if there are issues
make lint -> lint everything
make lint-yaml -> lint yaml
make lint-composer -> composer validate
make lint-xml -> lint xml
make lint-php -> php-cs-fixer

Those are for fixing the issues
make cs-fix
make cs-fix-php
make cs-fix-xml

make test
make docs
```

did I get it right?

jordisala1991:
This line is no longer needed and can be removed: https://github.com/sonata-project/dev-kit/blob/master/project/.travis.yml.twig#L45

greg0ire:
AFK, but yeah you got it right!

Soullivaneuh:
I really don't think we need a `.PHONY` for each task. You don't have any file or directory named `lint-yaml` on our project AFAIK.

greg0ire:
That's why you don't need to access the disk. And it is required for the `docs` target, at least.

Soullivaneuh:
> That's why you don't need to access the disk to check for the file.

Could you please elaborate?

> And it is required for the `docs` target, at least, because there is a `docs` directory.

Yeah, this is was it's already specified on current master.

greg0ire:
> Could you please elaborate?

This is best understood when running `make` in debug mode. It does way less checks when there are `.PHONY`, because of implicit rules it has: https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_10.html#SEC97 I don't remember how to reproduce this but that's something I experienced. Also, it's a good practice to always put `.PHONY` when you should because it might cause bugs if you don't notice there is a file / directory named like that, just like in `docs` case.

greg0ire:
More details about the performance implications here: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

Soullivaneuh:
Thank you @greg0ire

PR #383, opened 2018-01-23
[RFC] use build stages on travis

OskarStark:
refs https://docs.travis-ci.com/user/build-stages/#How-to-define-Build-Stages%3F

Proof PR: https://github.com/sonata-project/SonataNotificationBundle/pull/308

https://travis-ci.org/sonata-project/SonataNotificationBundle/builds/332209819?utm_source=github_status&utm_medium=notification

ohne titel

allowed failures are grouped under the stage, but an info note is shown IF a build failed which was allowed to fail!

### DIFF
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -42,17 +42,24 @@ env:
     - SYMFONY_DEPRECATIONS_HELPER=weak
     - TARGET=test
     - UPSTREAM_URL=https://github.com/sonata-project/{{ repository_name }}.git
+stages:
+  - lint
+  - docs
+  - test
 
-matrix:
+jobs:
   fast_finish: true
   include:
 {% if docs_target %}
-    - php: '{{ php|last }}'
+    - stage: docs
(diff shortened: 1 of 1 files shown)


greg0ire:
I tried that on greg0ire/enum, and then reverted it because this makes everything slower : you want to know upfront if you broke both the cs and the code

OskarStark:
> I tried that on greg0ire/enum, and then reverted it because this makes everything slower : you want to know upfront if you broke both the cs and the code

yes this is my other thought, but i wanted to try it and get my hands on, so I'm totally fine without merging it!

OskarStark:
ofc it will not execute Test-stage of lint fails for example

OskarStark:
and we should run a cron build on sandbox every day 😄 

screenshot 2018-01-23 13 57 51

greg0ire:
Yeah crons would be awesome too!

OskarStark:
No need for build stages

PR #385, opened 2018-01-24
Run lowest versions job together with lowest version of Symfony

covex-nn:
When i increased dependency version in my PR https://github.com/sonata-project/SonataFormatterBundle/pull/315 travis failed every time, i pushed to github. I tried to speed up composer to avoid failures. Here are my tries:

| Status | block-bundle | EnvVar | Job | Ran for |
| --- | --- | --- | --- | --- |
| Failure | ^3.11 | | [#939.7](https://travis-ci.org/sonata-project/SonataFormatterBundle/jobs/332731689) | 12 min 41 sec |
| Success | ^3.2 | | [#940.7](https://travis-ci.org/sonata-project/SonataFormatterBundle/jobs/332830146) | 11 min 48 sec |
| Success | ^3.11 | SYMFONY=2.8.0 | [#941.7](https://travis-ci.org/sonata-project/SonataFormatterBundle/jobs/332845215) | **4 min 39 sec** |
| Success | ^3.11 | SYMFONY_LTS=2 | [#942.7](https://travis-ci.org/sonata-project/SonataFormatterBundle/jobs/332945647) | 12 min 37 sec |
| Success | ^3.11 | SYMFONY=2.8.* | [#943.7](https://travis-ci.org/sonata-project/SonataFormatterBundle/jobs/333164319) | **4 min 53 sec** |

So, my suggestion is to pass `SYMFONY` env variable with lowest Symfony version to a job with `COMPOSER_FLAGS="--prefer-lowest"` variable:

```yaml
matrix:
  fast_finish: true
  include:
    # ...
    - php: '5.6'
      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY=2.8.*
    # ...
```

### DIFF
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -53,7 +53,7 @@ matrix:
     - php: '{{ php|last }}'
       env: TARGET=lint
     - php: '{{ php|first }}'
-      env: COMPOSER_FLAGS="--prefer-lowest"
+      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY={{ versions.symfony|first }}.*
 {% for package_name,package_versions in versions %}
 {% for version in package_versions %}
     - php: '{{ target_php|default(php|last) }}'


project/.travis.yml.twig
```diff
@@ -53,7 +53,7 @@ matrix:
     - php: '{{ php|last }}'
       env: TARGET=lint
     - php: '{{ php|first }}'
-      env: COMPOSER_FLAGS="--prefer-lowest"
+      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY={{ versions.symfony|first }}.0
```
dbu: as discussed, i agree with requiring the exact version for the lowest version. however, i am not sure if its a good idea to hardcode the `.0` - could we specify that in the config, or at least allow to override it? i have seen bundles that have e.g. symfony 2.8.18 as lowest version.

then again, if all sonata repos allow a .0 version in the lowest version, we can go with this and improve when we actually need it.

covex-nn: I've tried to run build with `SYMFONY=2.8.*` and updated a table at the top of a page.

A result is the same as with `SYMFONY=2.8.0`! So i think this is the best option!

covex-nn: Fixed

Soullivaneuh:
Bumping version support might be also an another solution to explore. There are some packages that can be easily upgraded on block-bundle like `sonata-project/cache` or `twig/twig`. And it is semver compliant.

covex-nn:
@Soullivaneuh my suggestion is just a compromise between execution time and lowness of dependencies. Yes, it is not 100% lowest. But it takes 4-5 travis minutes instead of >12, and travis will never break job execution.

greg0ire:
Yeah plus we do not support 2.4, do we? Should we ever find a bug due to it being installed, then we would not fix it and bump the version constraint. And since we should express all our deps, this shouldn't even happen.

dbu:
> Yeah plus we do not support 2.4, do we? Should we ever find a bug due to it being installed, then we would not fix it and bump the version constraint. 

we should bump all our dependencies to at least 2.8 imho. but in the future, similar things might happen when we think 3.3 is good enough, or with transient dependencies that are not directly ours. the problem is that we want to *know* that the bundle installed with the oldest dependencies will not break in random places. with the assumption of symfony/symfony on 2.8 we don't know, its only likely.

what is the problem if travis takes 12 minutes and we use travis_wait?

greg0ire:
> what is the problem if travis takes 12 minutes and we use travis_wait?

Slow builds and frustrated contributors?

covex-nn:
Sometimes travis_wait suddenly not found (i do not undestand why =) See failed jobs https://travis-ci.org/sonata-project/SonataFormatterBundle/builds/332717774

jordisala1991:
The only good part for long builds is that most of the times is related to something wrong on some dependency or the project itself. our last case wa the phpcr sonata bundle. It was using a package requiring symfony directly. Thanks to that long build we discovered the issue and fixed it. I do agree with @greg0ire tho, having builds randomly fail because of being long (without travis wait) is really frustrating

jordisala1991:
Travis wait can only be used on .travis.yml , not on any other sh file. That is probably why is failing

Soullivaneuh:
> Yeah plus we do not support 2.4, do we?

This is not the issue. My example is a case with a sub-dependencies, authorized to be installed. If we use this solution, we will never now if the project **really** works with the lowest dependencies.

As @dbu said in https://github.com/sonata-project/dev-kit/pull/385#issuecomment-360627688, this solution is dangerous. Maybe even making the `prefer-lowest` option completely useless.

> Slow builds and frustrated contributors?

Maybe. But a solution should be found on composer side, not on Travis configuration.

> Travis wait can only be used on .travis.yml , not on any other sh file. That is probably why is failing

Then, `travis_wait` on script call prefix?

jordisala1991:
This is fixed for formatterBundle already. Closing as it is not a great solution. If you find another build that fails because of timeout tell me to fix it

PR #389, opened 2018-02-04
Updated travis job for dashboard bundle

core23:
This bundle has no release, so we can bump symfony to symfony 3.4+

Refs https://github.com/sonata-project/SonataDashboardBundle/pull/131

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -122,7 +122,7 @@ dashboard-bundle:
     master:
       php: ['7.1', '7.2']
       versions:
-        symfony: ['2.8', '3.2', '3.3', '3.4']
+        symfony: ['3.4']
         sonata_core: ['3']
         sonata_admin: ['3']
         sonata_block: ['3']


jordisala1991:
2 questions:

- Can you try adding 4.0 to see if it passes?
- Can we agree on doing this change for all master branches? Stable 2.8, 3.2 (soon 3.3), 4.0 and Master 3.4 and 4.0. WDYT?

core23:
> Can you try adding 4.0 to see if it passes?

We don't need an explicit 4.0 version. Composer should do the magic by default

> Can we agree on doing this change for all master branches? Stable 2.8, 3.2 (soon 3.3), 4.0 and Master 3.4 and 4.0. WDYT?

Feel free to do this. I would like to have this one atomic, because it doesn't affect our stable builds.

jordisala1991:
We add versions of symfony here to enforce a build for that version.

About the other part I didnt mean to do it on this PR, just to have a path on what versions we would like to support on each branch

jordisala1991:
Sorry wrong button

core23:
> We add versions of symfony here to enforce a build for that version.

Only, if there is a minor version which isn't the latest one. E.g. if they release a 4.1 version, we should add 4.0 to our build. But for now, no action needed.

PR #393, opened 2018-02-18
Add xml lint to flintCI

jordisala1991:
With this one, flintCI becomes the same as we do with the travis lint build. So after this one, we could remove that build and make flintCI required on all projects.

@Soullivaneuh is the xml lint ready to use on our repos or is there something wrong with it?

### DIFF
--- a/project/.flintci.yml
+++ b/project/.flintci.yml
@@ -2,3 +2,4 @@ services:
   composernormalize: true
   phpcsfixer: true
   yamllint: true
+  xmllint: true


OskarStark:
@Soullivaneuh can you please give us feedback?

OskarStark:
> I don't think it should cause particular issue, but could you please make a proof PR for this one?

Here we go @Soullivaneuh https://github.com/sonata-project/SonataAdminBundle/pull/5296

If we merge this, we should remove the Travis XML Lint job, right?

OskarStark:
> If we merge this, we should remove the Travis XML Lint job, right?

Not sure if we should remove it, as it lints *.xliff, too.
I made a PR: https://github.com/sonata-project/dev-kit/pull/439

PR #396, opened 2018-02-23
Comment todo section

kunicmarko20:
Instead of always including To do section, we can comment it out, new contributors are often confused with this. I remember we discussed this on slack, wdyt?

### DIFF
--- a/project/.github/PULL_REQUEST_TEMPLATE.md
+++ b/project/.github/PULL_REQUEST_TEMPLATE.md
@@ -39,17 +39,17 @@ Closes #{put_issue_number_here}
 ### Security
 ```
 
-## To do
-
 <!--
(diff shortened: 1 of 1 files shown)


kunicmarko20:
Any chance we can get this merged, /cc @Soullivaneuh

Soullivaneuh:
At the reverse, it we left this commented, nobody will use it IMHO. :-/

greg0ire:
I'd rather have nobody use it than everybody misuse it.

kunicmarko20:
That is true,  they mostly just leave it and check everything. This way, people that know when to use it, can just uncomment it.

kunicmarko20:
So, can this get merged then? cc @Soullivaneuh

kunicmarko20:
ping @Soullivaneuh

PR #399, opened 2018-03-06
Try weak_vendors mode again

greg0ire:
This reverts commit e103e325b5ffec8fad180a5f3ddd0afc8322b72d.
The issues with this mode have been fixed and are released.


greg0ire:
The allowed failure, which is strict, stays. It will turn red if we are late on deprecations from our vendors.

Issue #390, opened 2018-02-05
Use Twig directly in sonata-project bundles without proxy classes, Stage 2


covex-nn:
Stage 2 goes after [Stage 1](https://github.com/sonata-project/dev-kit/issues/380), all changes must be pushed to master branch of each repository, because this is BC breaks.

There are 15 bundles, where were changes during Stage 1 and where `Sonata\BlockBundle\Templating\TwigEngine` is used now instead of `Symfony\Bundle\TwigBundle\TwigEngine`:

- [x] [BlockBundle](https://github.com/sonata-project/SonataBlockBundle/pull/476)
- [ ] [AdminBundle](https://github.com/sonata-project/SonataAdminBundle/pull/4938)
- [ ] SeoBundle
- [ ] PageBundle
- [ ] DoctrineORMAdminBundle
- [ ] DoctrinePhpcrAdminBundle
- [ ] FormatterBundle
- [ ] MediaBundle
- [ ] UserBundle
- [ ] NewsBundle
- [ ] ArticleBundle
- [ ] TranslationBundle
- [ ] TimelineBundle
- [ ] CommentBundle
- [ ] DashboardBundle

BlockBundle must be updated in accordance with the [plan](https://github.com/sonata-project/SonataBlockBundle/issues/456) or close to it.

For other bundles we should search for `EngineInterface` in `src` directory, change type hint to `Twig\Environment` and update tests. Also we should search for `templating` service in `docs` directory and change it to `twig`.

covex-nn:
Update `sonata-project/block-bundle` version inside "conflict" section of composer.json

- [ ] UserBundle
- [ ] SeoBundle
- [ ] MediaBundle
- [ ] CommentBundle
- [ ] ArticleBundle
- [ ] NewsBundle
- [ ] DoctrineORMAdminBundle

See https://github.com/sonata-project/dev-kit/issues/380#issuecomment-360392913

PR #400, opened 2018-03-15
Remove travis jobs with current stable BlockBundle for master branches

covex-nn:
After removing `symfony/templating` dependency from all bundles in `master` branch (see #390), packages won't be compatible with current stable version of `sonata-project/block-bundle`. Right now travis job with `env: SONATA_BLOCK=3.*` failing https://github.com/sonata-project/SonataAdminBundle/pull/4938 because of that incompatibility.

This PR removes jobs with `env: SONATA_BLOCK=3.*` and `env: SONATA_BLOCK='dev-master as 3.x-dev'` from `.travis-ci.yml`

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -5,7 +5,6 @@ admin-bundle:
       versions:
         symfony: ['3.4']
         sonata_core: ['3']
-        sonata_block: ['3']
     3.x:
       php: ['5.6', '7.0', '7.1', '7.2']
       versions:
@@ -125,7 +124,6 @@ dashboard-bundle:
         symfony: ['3.4']
         sonata_core: ['3']
         sonata_admin: ['3']
-        sonata_block: ['3']
 
 datagrid-bundle:
   branches:
@@ -185,7 +183,6 @@ doctrine-phpcr-admin-bundle:
       versions:
         symfony: ['3.4']
         sonata_admin: ['3']
-        sonata_block: ['3']
     2.x:
       php: ['5.6', '7.0', '7.1', '7.2']
       versions:
@@ -239,7 +236,6 @@ formatter-bundle:
       versions:
         symfony: ['3.4']
         sonata_core: ['3']
-        sonata_block: ['3']
     3.x:
       php: ['5.6', '7.0', '7.1', '7.2']
       versions:
@@ -327,7 +323,6 @@ page-bundle:
         symfony: ['3.4']
         sonata_core: ['3']
         sonata_admin: ['3']
-        sonata_block: ['3']
     3.x:
       php: ['5.6', '7.0', '7.1', '7.2']
       versions:
@@ -344,7 +339,6 @@ seo-bundle:
       php: ['7.1', '7.2']
       versions:
         symfony: ['3.4']
-        sonata_block: ['3']
         sonata_admin: ['3']
     2.x:
       php: ['5.6', '7.0', '7.1', '7.2']
@@ -361,7 +355,6 @@ timeline-bundle:
         symfony: ['3.4']
         sonata_core: ['3']
         sonata_admin: ['3']
-        sonata_block: ['3']
     3.x:
       php: ['5.6', '7.0', '7.1', '7.2']
       versions:


jordisala1991:
This should be only necessary for master branches right?

covex-nn:
Yes, these changes are made for `master` branches of AdminBundle, DashboardBundle, DoctrinePhpcrAdminBundle, FormatterBundle, PageBundle, SeoBundle and TimelineBundle.

I guess there will be a similar failures for `SONATA_ADMIN=3.x` too (if https://github.com/sonata-project/SonataAdminBundle/pull/4938 will be accepted), but i am not sure where it should be commented right now.

config/projects.yml
```diff
@@ -5,7 +5,7 @@ admin-bundle:
       versions:
         symfony: ['3.4']
         sonata_core: ['3']
-        sonata_block: ['3']
+        #sonata_block: ['3']
```
Soullivaneuh: Sorry, but I won't merge commented code/config.

If this one is not compatible with block-bundle 3, why not putting v4/dev-master?

covex-nn: When `sonata_block: ['3']` in present in `config/projects.yml` for some bundle, these lines are added to `.travis-ci.yml`:

```yaml
matrix:
  include:
    - php: '7.2'
      env: SONATA_BLOCK=3.*
    - php: '7.2'
      env: SONATA_BLOCK='dev-master as 3.x-dev'
  allow_failures:
    - env: SONATA_BLOCK='dev-master as 3.x-dev'
```

I cannot put `v4/dev-master` because after https://github.com/sonata-project/SonataAdminBundle/pull/4938 `master` branch of AdminBundle will become incompatible with current stable BlockBundle. And [travis job](https://travis-ci.org/sonata-project/SonataAdminBundle/jobs/357250777) for `SONATA_BLOCK=3.*` will always fail for `master` branch.

Also, all bundles, that uses BlockBundle are already use `3.x` version, so i guess there is no sence in additional job with `SONATA_BLOCK=3.*`

Also current stable branches of all bundles are already incompatible with `master` branch of BlockBundle and job with `SONATA_BLOCK='dev-master as 3.x-dev'` is always failing (but it is allowed).

So, with my PR i tried to remove jobs with `SONATA_BLOCK` env var

covex-nn:
@Soullivaneuh i reverted changes. Now i suggest to add jobs with SONATA_BLOCK env var to `matrix.allow_failures` block of `.travis-ci.yml`.

With my changes even if `master` branch of current project is not incompatible with `3.x` branch of BlockBundle, all travis build will not fail because of that.

project/.travis.yml.twig
```diff
@@ -70,6 +70,11 @@ matrix:
     - php: nightly
     - env: SYMFONY_DEPRECATIONS_HELPER=0
 {% for package_name,package_versions in versions %}
+{% if package_name|lower == 'sonata_block' %}
```
Soullivaneuh: The `projects.yml` file should be a simple configuration file.

I'm really uncomfortable with this hack, can't we just adjust the configuration file (without comment but remove if justified)?

covex-nn:
@Soullivaneuh i removed all lines with `sonata_block: ['3']` for master branches from `projects.yml`

config/projects.yml
```diff
@@ -5,7 +5,6 @@ admin-bundle:
       versions:
         symfony: ['3.4']
         sonata_core: ['3']
-        sonata_block: ['3']
```
Soullivaneuh: Okay now we can discuss on a diff. :stuck_out_tongue: 

Can't we just use `4`? Normally it should work with branch alias.

The thing that cause issue to me is to just remove sonata_block. We should test at least one version IMO.

covex-nn: A new version of AdminBundle is incompatible with current stable BlockBundle and can work only with a new version of BlockBundle, so i think it is not an option. I removed dependency on `^3.11` from composer.json and left only `^4.0@dev` for `sonata-project/block-bundle` dependency in my PR for AdminBundle, see diff https://github.com/sonata-project/SonataAdminBundle/pull/4938/files.

DashboardBundle, DoctrinePhpcrAdminBundle, FormatterBundle, PageBundle, SeoBundle and TimelineBundle has direct dependency on BlockBundle too. When i will update these bundles, i will update that dependency too

Soullivaneuh: @covex-nn Indeed, if the deps is already here, it will change nothing for now. But it would be better to keep it with only `4` in order to remember to test multiple versions in the future IMHO.

What do you think?

covex-nn: @Soullivaneuh a line with `sonata_block: ['4']` will generate two jobs for builds, see [.travis-ci.yml](https://github.com/sonata-project/SonataAdminBundle/blob/master/.travis.yml):

1. with `env: SONATA_BLOCK=4.*`
2. with `env: SONATA_BLOCK='dev-master as 4.x-dev'` (this job is allowed to fail)

These jobs first try to require another SonataBlockBundle (they take a version from SONATA_BLOCK env var) and then do all other stuff. See [before_install_test.sh.twig](https://github.com/sonata-project/dev-kit/blob/master/project/.travis/before_install_test.sh.twig)

BlockBundle v4 was not relesed yet. So, after updating `sonata_block: ['3']` to `sonata_block: ['4']` all builds will always fail because of job with `SONATA_BLOCK=4.*`.

I think, that such jobs should be added only for packages, that really has multiple versions. But there won't be multiple versions for SonataBlockBundle, there will be just `4`, because bundles will not be not compatible with lower versions =) Also, i think that all jobs `SONATA_BLOCK` and `SONATA_CORE` are useless now, because everything was already tested with regular jobs

covex-nn: @Soullivaneuh is there anything else that I could do here?

Soullivaneuh: @covex-nn Sorry for the time but I try to understand the need: Any project should work with at least one stable dependency. It's not the case here: https://github.com/sonata-project/SonataAdminBundle/pull/4938/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R29

If we have to release a new major of SonataAdminBundle, what should we do? At least, it should works with both versions or block bundle v4 should be release first.

Also, to test non-released dependencies, we already have the dev-master alias. But we may update the travis twig file to require 3.*@dev each time.

core23:
I think we can close this as @covex-nn is not that active anymore 😢

core23:
This is already fixed

Issue #173, opened 2016-06-16
Travis: Composer parallel installs with hirak/prestissimo plugin


Soullivaneuh:
https://packagist.org/packages/hirak/prestissimo

Try this plugin on a "big" project to see if it worth it.

Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1995

keradus:
just be sure to use at least ^0.2 if you use Composer ^1

Soullivaneuh:
> just be sure to use at least ^0.2 if you use Composer ^1

Thanks for the tip, we will use the latest one! :wink:

Soullivaneuh:
After a `composer clearcache` and deleted vendor on admin-bundle, I tried a `composer install`:

* Without the plugin: `1m17s`
* With the plugin: `0m29s`

This is indeed a great addition to have.

keradus:
You see, I know what am I doing ;)
Be aware that the boos is not that big when the deps is already cached (but then, you will have prestissimo cached as well, right? ;) )

Soullivaneuh:
> You see, I know what am I doing ;)

If you are talking about adding phpunit as a project dev, sorry but I still disagree. :stuck_out_tongue: 

> Be aware that the boos is not that big when the deps is already cached (but then, you will have prestissimo cached as well, right? ;) )

Yes we are caching composer files: https://github.com/sonata-project/dev-kit/blob/a4574f4ff3428bdd89ecf1c2f4f0190227abb250/.travis.yml#L13

But cache has a timeout, so...

jordisala1991:
I cant find where, but I know I already tested it on travis (and @greg0ire / you saw it) and we didnt gain any performance, due to the cache @keradus mentioned, and the fact that we are doing another command to install prestissimo in the first place.

Closing for now, feel free to reopen if this needs to be checked again.

PR #403, opened 2018-04-02
DX: Use prestissimo

keradus:
The gain won't be big, but I'm showing the approach how #173 shall be done.
If you want it - take it, if not ,close PR, no bad feelings !

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_install:
 install:
   - |
       composer global show hirak/prestissimo --quiet ||
-      travis_retry composer global require hirak/prestissimo --prefer-dist --no-interaction --no-progress $COMPOSER_FLAGS
+      travis_retry composer global require hirak/prestissimo --prefer-dist --no-interaction --no-progress
   - composer update --prefer-dist --no-interaction --no-progress $COMPOSER_FLAGS
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ sudo: false
 
 cache:
   directories:
-    - $HOME/.composer/cache/files
+    - $HOME/.composer
 
 env:
   global:
@@ -22,7 +22,10 @@ before_install:
   - wget http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar --output-document=php-cs-fixer.phar
 
 install:
-  - composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
+  - |
+      composer global show hirak/prestissimo --quiet ||
+      travis_retry composer global require hirak/prestissimo --prefer-dist --no-interaction --no-progress $COMPOSER_FLAGS
+  - composer update --prefer-dist --no-interaction --no-progress $COMPOSER_FLAGS
 
 script:
   - php php-cs-fixer.phar fix --verbose --dry-run --diff


.travis.yml
```diff
@@ -22,7 +22,8 @@ before_install:
   - wget http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar --output-document=php-cs-fixer.phar
 
 install:
-  - composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
+  - composer global show hirak/prestissimo -q || travis_retry composer global require hirak/prestissimo --prefer-dist --no-interaction --no-progress $COMPOSER_FLAGS
```
keradus: ref https://github.com/sonata-project/dev-kit/issues/173#issuecomment-377817983

if `prestissimo` is already installed, we don't need to install it again.
so same time as project dependencies are cached and `prestissimo` is not bringing that big benefits, we could cache it as well, so we won't lose any time if it's already there.

so with this change, we have 2 scenarios:
- no composer cache, install `prestissimo` plugin and then deps using plugin
- composer cache is there, don't install plugin and install deps from cache

jordisala1991: no need for composer flags here AFAIK, only for the composer update command

keradus: sure, I didn't found where that env var was set and what you provide as a value, so I kept it as is. dropping...

.travis.yml
```diff
@@ -22,7 +22,8 @@ before_install:
   - wget http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar --output-document=php-cs-fixer.phar
 
 install:
-  - composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
+  - composer global show hirak/prestissimo -q || travis_retry composer global require hirak/prestissimo --prefer-dist --no-interaction --no-progress
```
greg0ire: Please add some linebreaks here.

keradus: like between this and next `install` steps, or to split this line? not sure what you mean to be honest.
well, in the end I left allowance to modify my PR to you guys, so feel free to add them yourself however you want to have them!

greg0ire: Just to split this line, but yeah I can amend it.

.travis.yml
```diff
@@ -22,7 +22,10 @@ before_install:
   - wget http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar --output-document=php-cs-fixer.phar
 
 install:
-  - composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
+  - |
+      composer global show hirak/prestissimo --quiet ||
+      travis_retry composer global require hirak/prestissimo --prefer-dist --no-interaction --no-progress
+  - composer update --prefer-dist --no-interaction --no-progress $COMPOSER_FLAGS
```
jordisala1991: This composer flags is suspicious, I can't find it used anywhere :P

greg0ire: See 24cf39aa8f45364aa41f5681df508b7fe95fad9d

keradus: so same as me ;)
if you want to drop it, let us not mix the things and drop it in separate PR please

.travis.yml
```diff
@@ -10,7 +10,7 @@ sudo: false
 
 cache:
   directories:
-    - $HOME/.composer/cache/files
+    - $HOME/.composer
```
Soullivaneuh: Hmm, caching the whole composer directory looks dangerous to me. Why this is needed?

keradus: to have `prestissimo` cached. whole point of this PR.

I am caching whole composer for 3+ years, never had any issues because of that.
Why you find it dangeroues ?

Soullivaneuh: I find it dangerous because we might have not wanted cached file and I already add not wanted dependencies issues before.

But it was a long time ago, let's get a try.

Soullivaneuh:
Thanks!

.travis.yml
keradus: well, you can put anything in cache, what would be the outcome deps depends only on composer.* files, not caceh

PR #404, opened 2018-04-02
DX: PHP CS Fixer - don't use deprecated rule

keradus:


project/.php_cs.dist.twig
```diff
@@ -26,7 +26,7 @@ $rules = [
     'header_comment' => [
         'header' => $header,
     ],
-    'no_extra_consecutive_blank_lines' => true,
+    'no_extra_blank_lines' => true,
```
keradus: replaced with new one.

Soullivaneuh:
Thanks!

Issue #186, opened 2016-06-30
Use -i option of yaml-lint


Soullivaneuh:
In order to simplify the command.

Ref: https://github.com/Pryz/yaml-lint/issues/9#issuecomment-229568853

PR #405, opened 2018-04-03
Use -i and -e options of yaml-lint included in 0.0.10

aerostitch:
With the release of `yaml-lint` `0.0.10` it is now possible to exclude directories (in there, all the directories named `vendor` will be excluded and the non-yaml files will be ignored. The `-q` has been added to suppress non-useful output.

Closes: #186

### DIFF
--- a/project/Makefile
+++ b/project/Makefile
@@ -14,7 +14,7 @@ lint-composer:
 .PHONY: lint-composer
 
 lint-yaml:
-	find . -name '*.yml' -not -path './vendor/*' -not -path './src/Resources/public/vendor/*' | xargs yaml-lint
+	yaml-lint --ignore-non-yaml-files --quiet --exclude vendor .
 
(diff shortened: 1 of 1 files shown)


project/Makefile
```diff
@@ -14,7 +14,7 @@ lint-composer:
 .PHONY: lint-composer
 
 lint-yaml:
-	find . -name '*.yml' -not -path './vendor/*' -not -path './src/Resources/public/vendor/*' | xargs yaml-lint
+	yaml-lint -i -q -e vendor .
```
greg0ire: Are there long counterparts for these long options? If yes, it would be great to use them.

aerostitch: No problem. I just amended the commit to switch to their long options counterparts.

Soullivaneuh:
Nice improvement, thanks!

PR #408, opened 2018-04-13
Update LICENSE

kunicmarko20:
Not sure if we even need an end date.

### DIFF
--- a/project/LICENSE
+++ b/project/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2010-2018 Thomas Rabaix
+Copyright (c) 2010 Thomas Rabaix
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


project/LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2010-2016 Thomas Rabaix
+Copyright (c) 2010-2018 Thomas Rabaix
```
greg0ire: I think I read somewhere that you can just put "…2010 Thomas Rabaix" and be done with it forever

jordisala1991: o use twig to generate a date

kunicmarko20: https://github.com/symfony/form/commit/973a310098c590abe3e35a90ed2a7389f26158b3

was overridden with:

https://github.com/symfony/form/commit/0d89adcacf5be46c572d074302a0e0aa52e25fee

So I guess Symfony updates it every year, should we?

kunicmarko20: @greg0ire I saw that too but never find something that confirms it.

kunicmarko20:
So, 2018 or dropping end date? @sonata-project/contributors

greg0ire:
For me, dropping end date

kunicmarko20:
As @core23 also agrees and I don't really care, lets drop it then :smile:

kunicmarko20:
cc @Soullivaneuh

PR #413, opened 2018-05-26
Automate some of the release process

greg0ire:
This should make the process less boring/error-prone.
Can someone on Mac make it work for this OS and make a PR on this PR? There are probably some long options or some binaries not available `xdg-open` vs `open`?

### DIFF
--- a/latest-prs
+++ b/latest-prs
@@ -1,0 +1,34 @@
+#!/bin/bash -eu
+function stable_branch()
+{
+	git branch --format='%(refname:lstrip=2)'|grep '^[0-9].x$'|sort|tail --lines=1
+}
+
+function latest_tag_date()
+{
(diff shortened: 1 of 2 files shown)


Soullivaneuh:
> Can someone on Mac make it work for this OS and make a PR on this PR?

Can we simply have a Symfony command? You already have the GitHub API on it. :wink:

greg0ire:
It does not sound simple at all to me, but feel free to have a go at it.

greg0ire:
@Soullivaneuh why would we go through Symfony and php when all we actually want to do is run some bash commands? It sounds inefficient…

Soullivaneuh:
@greg0ire To centralise all on one command: dev-kit.

BTW, you use PHP here: https://github.com/sonata-project/dev-kit/pull/413/files#diff-5cac82c4d1c40e2bb8ddd40a3cc3524eR15

:stuck_out_tongue:

PR #424, opened 2018-09-22
Add docs to dev-kit

kunicmarko20:
Does this mean that docs config will now be handled by dev-kit? or we should make PR on all repos to fix docs?


project/docs/conf.py
```diff
@@ -0,0 +1,246 @@
...
+
+# The encoding of source files.
+#source_encoding = 'utf-8-sig'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'Sonata ~ AdminBundle'
+copyright = u'2010-2015, Thomas Rabaix'
```
OskarStark: we should change this to `2010-2018`

greg0ire: Or just 2010 IIRC

kunicmarko20: https://github.com/sonata-project/dev-kit/blob/master/project/LICENSE#L3 yep, just 2010, should I fix the style thing in a different PR?

OskarStark: No we can fix it here IMO

OskarStark:
@Soullivaneuh, please merge this

project/docs/conf.py
```diff
@@ -0,0 +1,246 @@
...
+source_suffix = '.rst'
+
+# The encoding of source files.
+#source_encoding = 'utf-8-sig'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'Sonata ~ AdminBundle'
```
OskarStark: Admin Bundle needs to be replaced by a template var imo

project/docs/Makefile
```diff
@@ -0,0 +1,153 @@
+# Makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line.
+SPHINXOPTS    =
+SPHINXBUILD   = sphinx-build
```
ElectricMaxxx: @kunicmarko20 can you add the env var usage here as i did in https://github.com/sonata-project/SonataSeoBundle/pull/282 pls doing so we can locally use a separate environment, which is common in python projects

kunicmarko20:
fixed.

OskarStark:
@Soullivaneuh please merge

Soullivaneuh:
I don't understand the goal here. The doc is generated by an another private repository.

kunicmarko20:
The goal is to make dev-kit handle shared files across repositories, in this case requirements.txt because sphinx 1.8.0 is broken and we need to skip it. If dev-kit can’t do this then we need to make PR on all repositories.

ElectricMaxxx:
@Soullivaneuh i really like it also, as it gives the oportunity to integrate a common testing proccess accross the packages for docs. The first think after a merge will be a pull on our cmf fork :-) cause i currently discuss to pull docs into single packages for [cmf also](https://github.com/symfony-cmf/symfony-cmf/issues/263) (discussion seems to become hard, but lets see)

kunicmarko20:
@Soullivaneuh can this work? or can I close it?

src/Console/Command/AutoMergeCommand.php
```diff
@@ -51,7 +51,7 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
     {
         parent::initialize($input, $output);
 
-        $this->projects = count($input->getArgument('projects'))
+        $this->projects = \count($input->getArgument('projects'))
```
Soullivaneuh: Not really related to the PR.

kunicmarko20: yes, but the build was failing

kunicmarko20: should I create another PR just for CS?

PR #427, opened 2018-10-06
Bump all php versions smaler then 7.1 in phpcr admin bundle

ElectricMaxxx:
As said in https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/521 i would have to drop some php version in Adminbundle

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -362,7 +362,7 @@ seo-bundle:
         sonata_block: ['3']
         sonata_admin: ['3']
     2.x:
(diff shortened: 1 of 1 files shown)


ElectricMaxxx:
@greg0ire pls give me a hint, when i have to rebase https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/521

greg0ire:
I don't think you will have to. This PR is only here to prevent dev-kit from adding old versions back, but it will not be able to if the build fails for those anyway :P

PR #432, opened 2018-11-10
Fix 7.3 apocalypse

greg0ire:
The last version is used as a default version for most jobs, and we do
not want to use 7.3 for this, since it is not stable yet. This should
fix the issue, and we can put 7.3 last when it is finally stable.


greg0ire:
We will just revert for now, until we find a better solution. Incoming PR from @OskarStark

PR #434, opened 2018-11-10
Revert "Run tests on php 7.3"

greg0ire:
This reverts commit 7fd5629248482e3c09c202806ee6b2cbfa993cb1.
This is not the right way to do it, because 7.3 is picked up as the
default for most jobs, and we do not want that at all.

### DIFF
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -67,7 +67,6 @@ matrix:
     - php: '{{ php|last }}'
       env: SYMFONY_DEPRECATIONS_HELPER=0
   allow_failures:
-    - php: 7.3
     - php: nightly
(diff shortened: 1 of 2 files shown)


project/.travis.yml.twig
```diff
@@ -67,7 +67,6 @@ matrix:
     - php: '{{ php|last }}'
       env: SYMFONY_DEPRECATIONS_HELPER=0
   allow_failures:
-    - php: 7.3
```
Soullivaneuh: Plus, didn't see it, but we don't need it. We already use nightly.

OskarStark:
Thank you @Soullivaneuh 👍

PR #440, opened 2018-11-13
changed chmod

OskarStark:
as mentioned by @dbu this file should be executable, too

screenshot 2018-11-13 11 12 50


greg0ire:
Hmmm wait… the twig file should not really be executable, only the produced sh file should. Are you sure this will work?

OskarStark:
Hmm 🤔 you are right, but it is also the only twig file with 644...

Let’s see if it works

dbu:
oh snap. okay, twig templates should not be executable. but there is another .twig file that already *is* executable, right? can we make the thing that compiles the .twig to an actual .sh do a chmod a+x on the generated files?

greg0ire:
Looks like this is how it is supposed to work:

https://github.com/sonata-project/dev-kit/blob/4d1fe369fac7a3e3f60498373d79300da8e1c463/src/Console/Command/DispatchCommand.php#L534

OskarStark:
> Shall we merge this then?

IIRC, it uses the CHMOD of the twig file, so YES

PR #445, opened 2018-11-27
Stop running the test suite on Symfony 2

greg0ire:
We should drop support for it. Also, if we get a bug with Symfony, it
will not get fixes on Symfony 2, and we cannot have that.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -10,7 +10,7 @@ admin-bundle:
     3.x:
       php: ['5.6', '7.0', '7.1', '7.2']
       versions:
-        symfony: ['2.8', '3.3', '3.4']
+        symfony: ['3.4']
         sonata_core: ['3']
         sonata_block: ['3']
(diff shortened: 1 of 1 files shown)


config/projects.yml
```diff
@@ -10,7 +10,7 @@ admin-bundle:
     3.x:
       php: ['5.6', '7.0', '7.1', '7.2']
       versions:
-        symfony: ['2.8', '3.3', '3.4']
+        symfony: ['3.3', '3.4']
```
core23: There is no active support for symfony 3.3 anymore: https://symfony.com/roadmap/3.3

OskarStark: Lets do it in a following PR, and see if everything will work here 😄

OskarStark:
Thank you!

config/projects.yml
OskarStark: already done in this PR 👍

Issue #372, opened 2017-12-12
Use composer plugin to load composer tools


core23:
Just found this sneaky [tool](https://github.com/bamarni/composer-bin-plugin) which could solve most of our problems when using php tools (e.g. php-cs-fixer, phpunit, ...).

Haven't tested it, but looks interesting:
https://medium.com/@tfidry/managing-your-dependencies-in-php-321d584441ab#e1ac

core23:
## Short status update

Works perfectly on my [projects](https://github.com/core23/LastFmBundle). You just need a `vendor-bin` with multiple `composer.json` files in it.

core23:
I've reopened this old issue, because it's a problem (again)

PR #446, opened 2018-11-29
Use a better strategy to pick the phpunit version

greg0ire:
Right now, repositories like the SonataArticleBundle one use phpunit 5.7
on their stable branch even though they only support php 7 and could use
phpunit 7. This moves the condition in the .travis.yml, meaning it will
be computed by Twig. We do not need bash conditions since we have one
.travis.yml per branch.

### DIFF
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -42,6 +42,7 @@ env:
     - SYMFONY_DEPRECATIONS_HELPER=weak_vendors
     - TARGET=test
     - UPSTREAM_URL=https://github.com/sonata-project/{{ repository_name }}.git
+    - PHPUNIT_VERSION={{ '5.6' in php ? '5.7': '7' }}
 
 matrix:
   fast_finish: true
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -42,7 +42,7 @@ env:
     - SYMFONY_DEPRECATIONS_HELPER=weak_vendors
     - TARGET=test
     - UPSTREAM_URL=https://github.com/sonata-project/{{ repository_name }}.git
-    - PHPUNIT_VERSION={{ '5.6' in php ? '5.7': '7' }}
+    - PHPUNIT_VERSION={{ '5.6' in php ? '5.7' : '7' }}
 
 matrix:
   fast_finish: true
--- a/project/.travis/install_test.sh
+++ b/project/.travis/install_test.sh
@@ -3,12 +3,6 @@ set -ev
 
 mkdir --parents "${HOME}/bin"
 
-# PHPUnit install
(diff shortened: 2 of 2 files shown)


greg0ire:
Disclaimer: I did not test any of this.

OskarStark:
I like it a lot 👍

core23:
We could also use a composer plugin to load dev-tools: https://github.com/sonata-project/dev-kit/issues/372

project/.travis.yml.twig
```diff
@@ -42,6 +42,7 @@ env:
     - SYMFONY_DEPRECATIONS_HELPER=weak_vendors
     - TARGET=test
     - UPSTREAM_URL=https://github.com/sonata-project/{{ repository_name }}.git
+    - PHPUNIT_VERSION={{ '5.6' in php ? '5.7', '7' }}
```
Soullivaneuh: I'm mitigated.

It is a nice improvement right now, but can be more complicated later depending to compatibility issues.

Soullivaneuh:
> Disclaimer: I did not test any of this.

shame

greg0ire:
I had no time, and I don't know how to besides maybe with twigfiddle :P

project/.travis.yml.twig
greg0ire: If it does, we may change in favor of explicit versions in the config, but until then, this is better, isn't it?

OskarStark: Yes it is

Soullivaneuh: `,` is wrong here, should be `:`.

https://twig.symfony.com/doc/2.x/templates.html#other-operators

OskarStark: Good catch 👍

greg0ire: Woops

greg0ire: Fixed, very nice catch!

OskarStark:
Can you please merge @Soullivaneuh ?

project/.travis.yml.twig
```diff
@@ -42,6 +42,7 @@ env:
     - SYMFONY_DEPRECATIONS_HELPER=weak_vendors
     - TARGET=test
     - UPSTREAM_URL=https://github.com/sonata-project/{{ repository_name }}.git
+    - PHPUNIT_VERSION={{ '5.6' in php ? '5.7': '7' }}
```
Soullivaneuh: Missing space between `'5.7'` and `:`.

OskarStark: added the missing space

greg0ire:
I don't know yet how to test. Would be great to have a dry run option for that

Soullivaneuh:
Indeed @greg0ire, sorry for that.

PR #447, opened 2018-11-29
Dx improvements

greg0ire:


greg0ire:
> Nice, does it work on MacOS now?

You tell me, I do not use that operating system :P

docker/php-console/entrypoint.sh
```diff
@@ -3,13 +3,4 @@ set -e
...
 setup
 
-# Wait for composer vendor
-if [ ! -f vendor/autoload.php ]; then
-    echo "Run 'composer install' command to continue."
-fi
-while [ ! -f vendor/autoload.php ]; do
-    sleep 1
-done
-echo "Composer vendor found, processing."
```
Soullivaneuh: Why removing this part?

greg0ire: Why would we wait before running `tail -f`? I think this was useful before, to tell people they need to run `composer` but now it works automatically

docker-compose.yml
```diff
@@ -31,5 +34,15 @@ console:
...
-  env_file:
-    - docker-compose.env
+  environment:
+    - UNIX_USERNAME
+    - UNIX_GROUP
+    - UNIX_UID
+    - UNIX_GID
+
+composer:
+  image: composer
```
Soullivaneuh: We already have composer and plugins we need on the php-dev image.

greg0ire: Yeah but how are you supposed to run `composer install`? It's not documented…

Soullivaneuh: `"echo "Run 'composer install' command to continue."`

On the script you deleted. :-)

I agree it's not very efficient, remember I did it very quickly at start. #DevExcuse :stuck_out_tongue:

docker/php-console/entrypoint.sh
```diff
@@ -3,13 +3,8 @@ set -e
 
 setup
 
-# Wait for composer vendor
 if [ ! -f vendor/autoload.php ]; then
-    echo "Run 'composer install' command to continue."
+	su "$(stat -c '%U' composer.json)" -c 'composer install'
```
greg0ire: @Soullivaneuh I reimplemented this, this time using the composer binary from the php-console image. Looks good?

docker-compose.yml
greg0ire: Fixed

docker/php-console/entrypoint.sh
```diff
@@ -3,13 +3,8 @@ set -e
 
 setup
 
-# Wait for composer vendor
 if [ ! -f vendor/autoload.php ]; then
-    echo "Run 'composer install' command to continue."
+	su "$UNIX_USERNAME" -c 'composer install'
```
greg0ire: @Soullivaneuh I reimplemented this, this time using the composer binary from the php-console image. Looks good?

Soullivaneuh: Let's try like that.

PR #449, opened 2018-12-02
Put subject first

greg0ire:
We can see a preview of the pull request in our Slack channel, but we
care most about the subject, not the rest. We can still check for the
other things first when reviewing the pull request.

# Before

before

# After

after


greg0ire:
Not that my "after" screenshot is inaccurate, because I deleted the ugly, all-caps comment before submitting my PR.

Soullivaneuh:
Nice idea, thanks!

PR #450, opened 2018-12-04
Use proper indention

greg0ire:
This caused many builds to fail because the stars are not aligned (the
one in the comments :P)


Soullivaneuh:
:star: :star: :star: :star: :star:

PR #452, opened 2018-12-06
add labels to issue templates

OskarStark:
see https://blog.github.com/changelog/2018-12-05-issue-template-automation-improvements/

### DIFF
--- a/project/.github/ISSUE_TEMPLATE/Bug.md
+++ b/project/.github/ISSUE_TEMPLATE/Bug.md
@@ -1,6 +1,7 @@
 ---
 name: 🐞 Bug Report
 about: Something is broken? 🔨
+labels: bug, unconfirmed
 ---
 
 <!--
--- a/project/.github/ISSUE_TEMPLATE/Feature.md
+++ b/project/.github/ISSUE_TEMPLATE/Feature.md
@@ -1,6 +1,7 @@
 ---
 name: 🚀 Feature Request
 about: I have a suggestion (and may want to implement it 🙂)!
+labels: feature
 ---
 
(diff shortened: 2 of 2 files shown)


project/.github/ISSUE_TEMPLATE/Bug.md
```diff
@@ -1,6 +1,10 @@
 ---
 name: 🐞 Bug Report
 about: Something is broken? 🔨
+title: ''
```
kunicmarko20: what is the point of title if it is empty?

project/.github/ISSUE_TEMPLATE/Bug.md
```diff
@@ -1,6 +1,10 @@
...
+labels: bug, unconfirmed
+assignees: ''
```
kunicmarko20: same question

project/.github/ISSUE_TEMPLATE/Bug.md
OskarStark: Its generated by the github ui in this way. You . can add a default title like:
`[BUG] `, but imo its to verbose when we have the bug label

greg0ire: You mean it breaks when it is missing?

greg0ire: Obviously not since we already use it that way, I'm dumb… so what's the point?

Soullivaneuh: I agree with @greg0ire. If the option is not mandatory I would recommend to not add it if not used.

PR #453, opened 2018-12-06
both libs have no docs

OskarStark:


OskarStark:
> This will change later? I mean we probably need to add some docs

I think so, but for now we don't want to have the docs build

PR #458, opened 2018-12-11
Initialize the phpcr repository when testing the SonataDoctrinePHPCRAdminBundle

ElectricMaxxx:
So we are able to separate scripting in .travis folder to have
customized solutions like we would like to introduce in
[#535](https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/535).


project/.travis/before_script_test.sh.twig
```diff
@@ -1,3 +1,12 @@
...
 
+{% if 'phpcr' in services %}
+TESTING_SCRIPTS_DIR=vendor/symfony-cmf/testing/bin
+CONSOLE=${TESTING_SCRIPTS_DIR}/console
+
+export KERNEL_CLASS="Sonata\\DoctrinePHPCRAdminBundle\\Tests\\Fixtures\\App\\Kernel"
+echo '+++ create PHPCR +++'
+${CONSOLE} doctrine:phpcr:init:dbal --drop --force -vvv
+${CONSOLE} doctrine:phpcr:repository:init -vvv
+{% endif %}
```
OskarStark: please add a newline

project/.travis/before_script_test.sh.twig
```diff
@@ -1,3 +1,12 @@
 #!/usr/bin/env bash
 set -ev
 
+{% if 'phpcr' in services %}
```
OskarStark: better check for project name here: doctrine-phpcr-admin-bundle

project/.travis/before_script_test.sh.twig
```diff
@@ -1,3 +1,12 @@
...
+TESTING_SCRIPTS_DIR=vendor/symfony-cmf/testing/bin
```
OskarStark: no need for this var IMO

ElectricMaxxx: i will check that.

project/.travis/before_script_test.sh.twig
ElectricMaxxx: according to the mongo service example you gave me, this would be consistent.

ElectricMaxxx: but having ci pipelines in mind, then services are additional software -  mostly containers, which is good for mongo, and here not :-1:

project/.travis/before_script_test.sh.twig
```diff
@@ -1,3 +1,11 @@
 #!/usr/bin/env bash
 set -ev
 
+{% if repository_name is same as('SonataDoctrinePhpcrAdminBundle') %}
```
OskarStark: do we have access to `repository_name` here @Soullivaneuh ?

ElectricMaxxx: more then on `package_name` :-) i called the `dev-kit` in `dry-run` localy (i know that thing as i cloned it for cmf)

OskarStark: Fine for me

dbu:
I edited the title.

project/.travis/before_script_test.sh.twig
```diff
@@ -1,3 +1,11 @@
...
+CONSOLE=vendor/symfony-cmf/testing/bin/console
+
+export KERNEL_CLASS="Sonata\\DoctrinePHPCRAdminBundle\\Tests\\Fixtures\\App\\Kernel"
+echo '+++ create PHPCR +++'
+${CONSOLE} doctrine:phpcr:init:dbal --drop --force -vvv
+${CONSOLE} doctrine:phpcr:repository:init -vvv
+{% endif %}
```
greg0ire: Wouldn't it be nicer to just have something along the lines of
```bash
if [ -f before_script_test.local.sh ]
then
    ./before_script_test.local.sh
fi
```
(that's probably not the correct syntax / doesn't work from any working directory but you get the idea)

dbu: i think that would be better to separate responsibilities between the repos, yes.

do we have such "hooks" anywhere else? we should be consistent with them, ideally.

greg0ire: I don't think we do.

Soullivaneuh: We must take care to not be tempted to write quickly some script that could be common.

But yeah, I would also prefer a *.local.sh method for that case.

Soullivaneuh: BTW, the @greg0ire method looks good to me.

Maybe place the condition on both places:

- Before the main script execution
- At the end of the main script execution

WDYT?

dbu: yeah, if we have before_script_test.local.sh we should likely also have after_script_test.local.sh :-D

OskarStark: I created a PR https://github.com/sonata-project/dev-kit/pull/465, shall I change sth. ?

PR #459, opened 2018-12-14
use concrete phpunit version as 5.7 is not available anymore

ElectricMaxxx:
Currently we do have several failing builds as fetching phpunit 5.7 ends in an 404. This PR fixes those builds by setting a concrete version for 5.7.x


ElectricMaxxx:
No idea. But i had several 404 while loading phpunit5.7.zip trough phpunit-bridge. And indeed the zip Archive or its link is gone.

PR #460, opened 2018-12-15
Add Travis configuration for exporter 2.x

greg0ire:


greg0ire:
Fair enough, let's do this. Anyway, we can always change our minds later.

PR #461, opened 2018-12-21
remove PHP-CS-Fixer from lint target

OskarStark:
we are using FlintCI and currently PHP-CS-Fixer does not support PHP 7.3

https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3697


OskarStark:
We should merge this **OR** #462

PR #462, opened 2018-12-29
use PHP 7.2 for linting, if 7.3 is available

OskarStark:
currently PHP-CS-Fixer does not support PHP 7.3 -> -> all DevKit PR's are broken -> all builds are failing


greg0ire:
Thanks @keradus !

PR #463, opened 2019-01-01
Stop testing unsupported versions of PHP

greg0ire:
Regardless of whether we continue to support php 5.6 or php 7.0, if a
new bug is discovered, and if affects these versions of php, and if
affects us, we will be stuck and will not be able to move forward. So
let us just stop testing with these versions and decide whether we stop
supporting them and when separately.

See https://secure.php.net/supported-versions.php

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -8,7 +8,7 @@ admin-bundle:
         sonata_block: ['3']
         symfony_maker: ['1.7']
     3.x:
-      php: ['5.6', '7.0', '7.1', '7.2', '7.3']
+      php: ['7.1', '7.2', '7.3']
       versions:
         symfony: ['3.4']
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
If we do this, we need to bump version on next minor of each package.

We can't just "don't test" on older version while continuing supporting it.

It will be a bug apparition nightmare. Simple example: Somebody using a PHP 7.1 only typing or function, test passes because we don't test lower... and boom. New issue.

greg0ire:
> If we do this, we need to bump version on next minor of each package.

Yes, let's do this!

Soullivaneuh:
@greg0ire Maybe could you open a dev-kit issue with the package list? Or maybe an another support to follow the version bumping ?

PR #467, opened 2019-01-09
Handle void return manually

greg0ire:
This is an emergency measure, maybe we will be able to re-enable it in
the future, or enable it conditionally.

I noticed things could be simplified in that file due to dropping php 5, but did not because I don't want this one to be blocked.


kunicmarko20:
But this is master, why is this a problem here?

PR #471, opened 2019-01-15
Implement xdebug_set_filter

greg0ire:
This allows us to ignore vendor directories and phpunit code when
producing code coverage, making everything faster

https://thephp.cc/news/2019/01/faster-code-coverage


greg0ire:
Proof PR: https://github.com/sonata-project/SonataAdminBundle/pull/5428

project/Makefile
```diff
@@ -51,8 +51,14 @@ cs-fix-xml:
 	done
 .PHONY: cs-fix-xml
 
-test:
-	phpunit -c phpunit.xml.dist --coverage-clover build/logs/clover.xml
+build:
+	mkdir $@
+
+build/xdebug-filter.php: phpunit.xml.dist build
```
Soullivaneuh: Don't we need a clean method for this file?

greg0ire: If we didn't for `build/logs/clover.xml`, then we do not for that one. And we probably do not because `build` is gitignore'd . I can implement that if you still feel the need for it.

PR #476, opened 2019-01-19
use local test script from .travis folder

ElectricMaxxx:
As orderd in https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/521#discussion_r248939425 

Todo:
* [ ] see if the change works in that PR mentioned above


ElectricMaxxx:
Yes it work on that PR. so it can be merged

PR #495, opened 2019-05-01
Phpunit per branch

greg0ire:
Please read the commit messages.

@soullivaneuh do you think it should work? How can I test this?

### DIFF
--- a/src/Config/ProjectsConfiguration.php
+++ b/src/Config/ProjectsConfiguration.php
@@ -60,6 +60,7 @@ class ProjectsConfiguration implements ConfigurationInterface
                                         ->arrayNode('php')->prototype('scalar')->defaultValue([])->end()->end()
                                         ->arrayNode('services')->prototype('scalar')->defaultValue([])->end()->end()
                                         ->scalarNode('target_php')->defaultNull()->end()
+                                        ->scalarNode('phpunit_version')->defaultNull()->end()
                                         ->append($this->addVersionsNode())
                                         ->scalarNode('docs_path')->defaultValue('docs')->end()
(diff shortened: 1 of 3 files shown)


greg0ire:
> How can I test this?

Oh there is a dry run in Travis, nice! But we've reached the hourly limit :P

src/Config/ProjectsConfiguration.php
```diff
@@ -60,6 +60,7 @@ public function getConfigTreeBuilder()
                                         ->arrayNode('php')->prototype('scalar')->defaultValue([])->end()->end()
                                         ->arrayNode('services')->prototype('scalar')->defaultValue([])->end()->end()
                                         ->scalarNode('target_php')->defaultNull()->end()
+                                        ->scalarNode('phpunit_version')->defaultNull()->end()
```
OskarStark: Why not default 8 here? And no default in the twig template?

greg0ire: I noticed this file afterwards, am on my phone now so can't change now, might change later if doing so makes more sense (I'm not sure, will think about it)

greg0ire: @OskarStark I gave this a second thought, and we might have to reintroduce conditions as above later, ending up with something like this: `{{ phpunit_version|default('7.1' in php ? '7' : '8')}}`. If we put the default in the config, then we would not know from twig if what is specified is the default value or something we explicitely set, meaning we would have no way to override that kind of logic I think we might reintroduce. Can we keep things as is?

OskarStark: That makes sense 👍🏻
Sure we keep it like this

@Soullivaneuh please merge

PR #497, opened 2019-05-07
Drop php 7.1 from datagrid 3.x

greg0ire:
It was never supposed to be supported.


core23:
Only for the datagrid bundle?

greg0ire:
Has to do with https://github.com/sonata-project/SonataDatagridBundle/pull/188, are there any other offenders?

greg0ire:
Ah right it's in security mode… I might update this PR later to drop it from every lib… has to be thought through though.

PR #498, opened 2019-05-16
Use phpunit 8 for the exporter lib

greg0ire:


config/projects.yml
```diff
@@ -233,12 +233,14 @@ exporter:
         symfony: ['3.4']
         doctrine_odm: ['1']
       services: [mongodb]
+      phpunit_version: 8
```
phansys: [8.1](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-8.1.md) and 8.2 ship some interesting fixes. Could we use 8.2 as constraint?

greg0ire: That's not a constraint, that's just part of a url. Right now, if you download https://phar.phpunit.de/phpunit-8.phar, you will get 8.2.4.

PR #503, opened 2019-07-08
Use PHPUnit bridge from own dependencies

phansys:
Replaces #265.
See sonata-project/SonataAdminBundle#5617 for some context.

**TODO**:
- [x] Fix performance issues.


soullivaneuh:
Can't you just use a `.phar` to get code completion? Or I am missing something?

If we could avoid to require the whole phpunit stack (plus the locked shipped version and deps we will lose), it would be better.

phansys:
This is not for autocompletion, its about consistency between the tool used by the CI suite and the one used locally by the users. AFAIK, the deps will not be lost, since PHPUnitBridge installs their dependencies in a separate `vendor` dir.
BTW, thinking about the way users execute the tests, we should document that the `SYMFONY_PHPUNIT_REMOVE` env var must be set before running the command, in order to avoid "phpspec/prophecy" to be removed:
```
SYMFONY_PHPUNIT_REMOVE=symfony/yaml vendor/bin/simple-phpunit
```

OskarStark:
Do we still need this ? My feelings about the configurable Phpunit Version are quite good at the moment

VincentLanglet:
@phansys You should set the phpunit version too cf https://github.com/sonata-project/SonataAdminBundle/pull/5939

And we should have some documentation about how to run tests in the different README.
I was just running `vendor/bin/simple-phpunit` before, and I had always failing tests (That's why I made the PR #5939).

project/Makefile
```diff
@@ -58,14 +58,14 @@ HAS_XDEBUG=$(shell php --modules|grep --quiet xdebug;echo $$?)
 
 build/xdebug-filter.php: phpunit.xml.dist build
 ifeq ($(HAS_XDEBUG), 0)
-	phpunit --dump-xdebug-filter $@
+	SYMFONY_PHPUNIT_REMOVE=symfony/yaml vendor/bin/simple-phpunit --dump-xdebug-filter $@
```
franmomu:
```suggestion
	SYMFONY_PHPUNIT_VERSION=8.5.0 SYMFONY_PHPUNIT_REMOVE=symfony/yaml vendor/bin/simple-phpunit --dump-xdebug-filter $@
```

phansys: Thank you!

phansys:
Thank you @franmomu. I've updated the PR based on your comments.

project/.travis.yml.twig
```diff
@@ -33,9 +33,10 @@ env:
   global:
     - PATH="$HOME/.local/bin:$PATH"
     - SYMFONY_DEPRECATIONS_HELPER=max[self]=0
+    - SYMFONY_PHPUNIT_VERSION={{ phpunit_version }}
+    - SYMFONY_PHPUNIT_REMOVE=symfony/yaml
```
greg0ire: Let's take this an occasion to use alpha sort
```suggestion
    - SYMFONY_PHPUNIT_REMOVE=symfony/yaml
    - SYMFONY_PHPUNIT_VERSION={{ phpunit_version }}
```

src/Config/ProjectsConfiguration.php
```diff
@@ -55,7 +55,7 @@ public function getConfigTreeBuilder()
                                         ->arrayNode('php')->prototype('scalar')->defaultValue([])->end()->end()
                                         ->arrayNode('services')->prototype('scalar')->defaultValue([])->end()->end()
                                         ->scalarNode('target_php')->defaultNull()->end()
-                                        ->scalarNode('phpunit_version')->defaultValue('8')->end()
+                                        ->scalarNode('phpunit_version')->defaultValue('8.5.0')->end()
```
greg0ire: Shouldn't this be removed entirely? :thinking:

phansys: Done.

phansys:
> I think we should either put relevant variables in `.travis` or in `phpunit.xml`, but not in both. See [sonata-project/SonataAdminBundle#5939](https://github.com/sonata-project/SonataAdminBundle/pull/5939)

Then, I think we should provide them at `phpunit.xml.dist` in order to make sure every local clone of our projects has these variables available. WDYT?

franmomu:
> > I think we should either put relevant variables in `.travis` or in `phpunit.xml`, but not in both. See [sonata-project/SonataAdminBundle#5939](https://github.com/sonata-project/SonataAdminBundle/pull/5939)
> 
> Then, I think we should provide them at `phpunit.xml.dist` in order to make sure every local clone of our projects has these variables available. WDYT?

It makes sense 👍 to me in `phpunit.xml.dist`

greg0ire:
Can you please squash your commits? I don't think it makes sense to have the variables appear in one file only to immediately disappear in the next commit.

project/.travis.yml.twig
```diff
@@ -35,7 +35,6 @@ env:
     - SYMFONY_DEPRECATIONS_HELPER=max[self]=0
```
franmomu: Should we move this as well? 🤔

greg0ire: Yes we should, we don't want the build to fail on Travis and not locally. env vars defined in `phpunit.xml` can be overriden from the CLI, right?

phansys: I didn't move this one because below in this file, there are some conditionals which set it to `0`.
Regarding the CLI overriding, I need to check that. I remember [some issues](https://github.com/symfony/symfony/issues?q=is%3Aissue+SYMFONY_DEPRECATIONS_HELPER) about this behavior but I haven't them currently in mind.

phansys: I've checked this locally, seems to be working fine.
I've also removed `SymfonyTestsListener` from `phpunit.xml.dist` since this is automatically registered when using the bridge.

greg0ire:
I fixed the conflicts :)

OskarStark:
One question

phansys:
AFAIK, there is no need for changes on `projects.yml`. Maybe we should check later projects where `phpunit.xml.dist` is excluded.

OskarStark:
IIRC we had some custom phpunit_version set ?

greg0ire:
@core23 had me remove it in this PR: https://github.com/sonata-project/dev-kit/pull/541#discussion_r372223397

PR #504, opened 2019-07-19
Configure "ensure_fully_multiline" rule at `.php_cs.dist.twig`

phansys:
For a sample diff, check https://github.com/sonata-project/SonataAdminBundle/compare/3.x...phansys:method_argument_space?expand=1.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -26,6 +26,7 @@ $rules = [
     'header_comment' => [
         'header' => $header,
     ],
+    'method_argument_space' => ['on_multiline' => 'ensure_fully_multiline'],
     'no_extra_blank_lines' => true,
     'no_php4_constructor' => true,
     'no_useless_else' => true,
--- a/project/.php_cs.dist.twig
+++ b/project/.php_cs.dist.twig
@@ -26,6 +26,7 @@ $rules = [
     'header_comment' => [
         'header' => $header,
     ],
+    'method_argument_space' => ['on_multiline' => 'ensure_fully_multiline'],
     'no_extra_blank_lines' => true,
     'no_php4_constructor' => true,
     'no_superfluous_phpdoc_tags' => ['allow_mixed' => true],
--- a/src/Command/DispatchCommand.php
+++ b/src/Command/DispatchCommand.php
@@ -376,8 +376,11 @@ final class DispatchCommand extends AbstractNeedApplyCommand
 
             // If the previous branch is not merged into the current one, do nothing.
             if ($previousBranch && $this->githubClient->repos()->commits()->compare(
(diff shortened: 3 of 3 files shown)


src/Console/Command/DispatchCommand.php
```diff
@@ -373,7 +373,10 @@ private function dispatchFiles(Package $package): void
             }
             // If the previous branch is not merged into the current one, do nothing.
             if ($previousBranch && $this->githubClient->repos()->commits()->compare(
-                    static::GITHUB_GROUP, $repositoryName, $currentBranch, $previousBranch
+                static::GITHUB_GROUP,
+                $repositoryName,
+                $currentBranch,
+                $previousBranch
                 )['ahead_by']) {
```
greg0ire: There is an indention issue here.

OskarStark: If I am correct the indention is not checked by this rule automatically.

Anyway I like the result but we should check this after the dev kit updates and FlintCI fix PRs

greg0ire: Yes, you're correct IIRC, we will have to check. Luckily, people seldom violate that rule.

soullivaneuh: In any case, there is an indentation issue here. No? :-)

OskarStark: @phansys can you please fix this?

phansys: The proposed fix doesn't look right to me (https://flintci.io/repositories/50/analyses/24032):
```diff
@@ -373,11 +373,11 @@ final class DispatchCommand extends AbstractNeedApplyCommand
             }
             // If the previous branch is not merged into the current one, do nothing.
             if ($previousBranch && $this->githubClient->repos()->commits()->compare(
-                    static::GITHUB_GROUP,
-                    $repositoryName,
-                    $currentBranch,
-                    $previousBranch
-                )['ahead_by']
+                static::GITHUB_GROUP,
+                $repositoryName,
+                $currentBranch,
+                $previousBranch
+            )['ahead_by']
             ) {
                 continue;
             }
```
Maybe we should squash this call into a single line.

OskarStark: Can you post an issue to the PHP-CS-Fixer guys? Maybe it gets fixed by another release? 

@Soullivaneuh which version is FlintCI using?

OskarStark:
Shall we proceed here and check the results afterwards in the DevKit PR’s or close entirely?

greg0ire:
I would proceed… the cases will only be edge cases IMO.

OskarStark:
Me too 👍🏻

PR #506, opened 2019-08-13
Added support for vendor-bin tools

### LINKED ISSUE #505: Configure external tools
We should add some external tools to the config that are needed for the build pipeline (e.g. phpunit, phpstan, php-cs-fixer, ...), so we can use the [composer-bin plugin](https://github.com/sonata-project/dev-kit/issues/372).

Here is an example config:
```yaml
block-bundle:
  branches:
    master:
      php: ['7.3']
      tools:
          - phpstan
          - phpunit
      versions:
        symfony: ['4.3']
```

This will add new entries to `.travis.yml` and `Makefile`.

We need this for this PR: https://github.com/sonata-project/SonataBlockBundle/pull/611

core23:
Fixes #505

### DIFF
--- a/project/.travis/install_common.sh
+++ b/project/.travis/install_common.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+set -ev
+
+composer update --prefer-dist --no-interaction --prefer-stable ${COMPOSER_FLAGS}
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -47,6 +47,8 @@ block-bundle:
       php: ['7.3']
       versions:
         symfony: ['4.3']
+      tools:
+        - phpstan
     3.x:
       php: ['7.1', '7.2', '7.3']
       versions:
--- a/project/.travis.yml.twig
+++ b/project/.travis.yml.twig
@@ -82,7 +82,7 @@ before_install:
   - if [ -x .travis/before_install_${TARGET}.sh ]; then .travis/before_install_${TARGET}.sh; fi;
 
 install:
-  - if [ -x .travis/install_${TARGET}.sh ]; then .travis/install_${TARGET}.sh; fi;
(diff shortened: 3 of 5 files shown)


project/.travis/install_common.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+set -ev
+
+composer update --prefer-dist --no-interaction --prefer-stable ${COMPOSER_FLAGS}
```
soullivaneuh: AFAIK, it will update the project dependency. What is the point with external tools?

It does not seems very clear IMHO.

core23: The `vendor-bin` plugin uses the normal composer build and registers a post update hook which will then load the composer-bin dependencies, so we need a normal composer install/update first.

https://github.com/bamarni/composer-bin-plugin

core23:
>     1. Why do not use phpstan as default for each project, like we do for phpunit?

We can do this, but we need an executable on each repo

>     2. Why not using phar binaries directly?

Because, we could not use phpstan extensions with the phar build

core23:
ping @soullivaneuh

project/.travis/install_common.sh
soullivaneuh: You need the `bin` suffix:

```
composer bin update
```

https://github.com/bamarni/composer-bin-plugin#usage

Am I right?

core23: Yes, but you need at least the plugin itself, which is installed by the *normal* composer

OskarStark:
GithubActions is now out of BETA, so we can use it instead.

Is it only once with the latest php version?

core23:
I'm giving up 😞 

The idea was to provide the same tools for your local build **AND** github build. GA are just a solution for the CI build.

OskarStark:
No, if you add a composer script which gets executed by the GA, you can locally use:
`composer phpstan` too

PR #524, opened 2019-11-28
Move PHP version range to 7.2 - 7.4

core23:
PHP 7.1 support will end in 2 days.
PHP 7.4 was released today


OskarStark:
Afterwards we need to adjust all composer min versions 😟
Still 👍🏻 for monorepo like Symfony

OskarStark:
Can you please check Travis?

PR #530, opened 2019-12-12
Add support for Probot Stale

### LINKED ISSUE #529: Use stale bot for inactive issues
We have many issues that are long running / did not get any attention. 

AFAIK we should close them after a while. If it's still a problem for anyone, it should come back or someone should be working on this.

[Stale](https://probot.github.io/apps/stale/) could close this issues for us, so we can focus on active issues.

phansys:
Closes #529.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,11 @@
+daysUntilStale: 180
+daysUntilClose: 7
+exemptLabels:
+  - keep
+  - critical
+staleLabel: pending author
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+closeComment: true
--- a/project/.github/stale.yml
+++ b/project/.github/stale.yml
@@ -1,0 +1,1 @@
+_extends: dev-kit
--- a/config/dev-kit.yml
+++ b/config/dev-kit.yml
@@ -41,6 +41,8 @@ labels:
     color: '006b75'
   - name: 'hacktoberfest'
     color: 'b0581d'
+  - name: 'keep'
+    color: '59b3c1'
 
 packages:
   symfony: symfony/symfony


project/.github/stale.yml
```diff
@@ -0,0 +1,16 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
```
OskarStark: excePt?

project/.github/stale.yml
```diff
@@ -0,0 +1,16 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
```
core23: Keep in mind, this will close nearly all issues...

core23: Maybe we should start with a higher value first... 180 days?

OskarStark: Yes why not

core23:
We should keep one config inside the `dev-kit` and use `_extends` in the other repositories:

https://probot.github.io/docs/best-practices/#store-configuration-in-the-repository

project/.github/stale.yml
phansys: Nope, this is "[exempt](https://github.com/probot/stale/blob/cd1e7623c6971731d4c19846d62b4dd383616171/lib/stale.js#L68)".

core23:
@soullivaneuh Can you merge this and install stale for us? https://probot.github.io/apps/stale/

.github/stale.yml
```diff
@@ -0,0 +1,16 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 180
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - critical
```
core23: Should we add an other label like `no stale` to keep specific issues open?

phansys: Yes, or `wont stale` (to be consistent with `wont fix`).
Let me know and I'll be pushing again.

core23: Or something like `keep open` or just `keep`

phansys: Done!

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 180
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - keep
```
core23: You should also update this file: https://github.com/sonata-project/dev-kit/blob/master/config/dev-kit.yml

phansys: It's updated now.

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
```
soullivaneuh: The configuration labels are enough explicit, I think the comments can be removed from this file.

phansys: It's ok. The comments were removed.

OskarStark:
Is this FOR devkit or for the projects?

core23:
> Is this FOR devkit or for the projects?

both

core23:
After merging this, we only got ~150 open issues :rocket: 

https://github.com/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+org%3Asonata-project+archived%3Afalse+updated%3A%3E2019-06-01+

OskarStark:
Ok but in the repo we end up with a file which extends something which isn’t in the repo right? Only files in /project dir will be added by a PR

core23:
> Ok but in the repo we end up with a file which extends something which isn’t in the repo right? Only files in /project dir will be added by a PR

That's right the `dev-kit` will put a simple config inside `all repos` and that file will extend the file located inside the `dev-kit` repo.

phansys:
> Ok but in the repo we end up with a file which extends something which isn’t in the repo right? Only files in /project dir will be added by a PR

Yes, based on the leveraging of the feature proposed here: https://github.com/sonata-project/dev-kit/pull/530#issuecomment-564880897.

>Just put common configuration keys in a repository within your organization. Then reference this repository from config files with the same name.

PR #535, opened 2019-12-21
Cleanup admin bundle tests

core23:
We don't need these tests anymore

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -4,9 +4,7 @@ admin-bundle:
       php: ['7.2', '7.3', '7.4']
       target_php: '7.3'
       versions:
-        symfony: ['3.4']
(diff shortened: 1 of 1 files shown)


config/projects.yml
```diff
@@ -4,9 +4,7 @@ admin-bundle:
       php: ['7.2', '7.3', '7.4']
       target_php: '7.3'
       versions:
-        symfony: ['3.4']
-        sonata_core: ['3']
-        sonata_block: ['3']
+        symfony: ['4.2', '4.3', '4.4', '5.0']
```
OskarStark: Why 4.2 ?

core23: Changed this to 4.4+

PR #539, opened 2020-01-14
Drop testing php 7.2 in master branch of SonataAdminBundle

franmomu:
It will [use SonataBlockBundle which the requirement is `^7.3`](https://github.com/sonata-project/SonataBlockBundle/blob/4.x/composer.json#L23)

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -1,7 +1,7 @@
 admin-bundle:
   branches:
     master:
-      php: ['7.2', '7.3', '7.4']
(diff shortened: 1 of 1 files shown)


OskarStark:
So we need to increase php in composer.json of admin bundle to 7.3?

franmomu:
> So we need to increase php in composer.json of admin bundle to 7.3?

I forgot to link the PR 😬 https://github.com/sonata-project/SonataAdminBundle/pull/5833

PR #541, opened 2020-01-27
Use PHPUnit 8 for all packages

greg0ire:
Some proof PRs:

- https://github.com/sonata-project/SonataPageBundle/pull/1105
- https://github.com/sonata-project/sonata-doctrine-extensions/pull/164
- https://github.com/sonata-project/SonataNotificationBundle/pull/408

PR #540, opened 2020-01-22
Bump LTS version number

greg0ire:
https://github.com/sonata-project/SonataAdminBundle/pull/5843

PR #542, opened 2020-01-28
Use PHPUnit 8 for the notification bundle

bmaziere:
Related to https://github.com/sonata-project/SonataNotificationBundle/pull/408


core23:
Thanks @bmaziere 

We will do this for every project #540 #541

PR #550, opened 2020-02-03
Add wontfix label for Stale-Bot

OskarStark:
If not needed, please close


core23:
We don't need this. The stale app adds this. The cron uses the right label.

OskarStark:
Thanks for the feedback!

jordisala1991:
But, if you don't add this, SonataCI will remove the label all the time.
@core23

PR #581, opened 2020-02-12
Workflows

OskarStark:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,32 @@
+on: [push, pull_request]
```
OskarStark: I renamed the file from `ci.yaml` to `lint.yaml` to be the same like in the `/project` folder

PR #599, opened 2020-02-25
Test Symfony 5.0

OskarStark:
What do you think?


core23:
This won't work for most bundles, because the last 3.x stable release of the block bundle does not support symfony 5.

PR #602, opened 2020-02-27
Add PHPStan

franmomu:
Add PHPStan check to all projects

Ref: https://github.com/sonata-project/sonata-doctrine-extensions/pull/178#issuecomment-591625037

### DIFF
--- a/project/.github/workflows/qa.yaml
+++ b/project/.github/workflows/qa.yaml
@@ -1,0 +1,26 @@
+on:
+    push:
+        paths:
+            - 'src/**'
+            - 'phpstan.neon.dist'
+
+    pull_request:
+        paths:
+            - 'src/**'
+            - 'phpstan.neon.dist'
+
+name: Quality assurance
+jobs:
+    phpstan:
+        name: PHPStan
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v2
+
+            - name: PHPStan
+              uses: "docker://oskarstark/phpstan-ga"
+              env:
+                  REQUIRE_DEV: 'true'
+              with:
+                  args: analyse
--- a/project/phpstan.neon.dist
+++ b/project/phpstan.neon.dist
@@ -1,0 +1,5 @@
+parameters:
+    level: 0
+
+    paths:
+        - src


OskarStark:
Please add 2 files.

`DELETE_qa.yml` to delete this file if available int he repo and a `qa.yaml` which will then result in a rename from: `qa.yml` -> `qa.yaml`

franmomu:
> We need a default `phpstan.neon.dist` config for this

I don't know well `dev-kit` yet, is there a way to add it if it does not exist?

OskarStark:
just add it to the `/projects` folder here with the name it should have :)

OskarStark:
Wen should only trigger this if:
https://github.com/sonata-project/SonataAdminBundle/blob/db7ea29c36fb0673e29ad2eaebead0e18e6c1198/.github/workflows/documentation.yaml#L1-L8

+

project/phpstan.neon.dist
```diff
@@ -0,0 +1,5 @@
+parameters:
```
core23: Will this overwrite the existing file?

OskarStark: you mean in admin bundle for example? Yes

core23: Then we should find a way to create files via dev-kit, if they don't exist

franmomu: If we are using the `DELETE_` prefix, what about `DEFAULT_filename.extension`?

project/.github/workflows/qa.yaml
```diff
@@ -0,0 +1,26 @@
...
+            - 'phpstan.neon.dist'
+
+    pull_request:
+        paths:
+            - 'src/**'
+            - 'phpstan.neon.dist'
+
+name: Quality assurance
+jobs:
+  phpstan:
```
FlintCIBot: - yamllint: wrong indentation: expected 4 but found 2 (indentation)

project/.github/workflows/qa.yaml
```diff
@@ -0,0 +1,26 @@
...
+    name: PHPStan
```
FlintCIBot: - yamllint: wrong indentation: expected 6 but found 4 (indentation)

project/.github/workflows/qa.yaml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
```
FlintCIBot: - yamllint: wrong indentation: expected 8 but found 6 (indentation)

project/.github/workflows/qa.yaml
```diff
@@ -0,0 +1,26 @@
...
+        uses: actions/checkout@v2
+
+      - name: PHPStan
+        uses: "docker://oskarstark/phpstan-ga"
+        env:
+          REQUIRE_DEV: 'true'
```
FlintCIBot: - yamllint: wrong indentation: expected 12 but found 10 (indentation)

project/.github/workflows/qa.yaml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          args: analyse
```
FlintCIBot: - yamllint: wrong indentation: expected 12 but found 10 (indentation)

project/phpstan.neon.dist
OskarStark: What do you mean? Can you make an example?

franmomu: I meant to do something similar to the `DELETE_` prefix, so instead of pushing the `qa.yaml`, pushing `DEFAULT_qa.yaml`. So if the `qa.yaml` exists does nothing, and if it doesn't, it will create the `qa.yaml`.

OskarStark: Ok but in this case we would not get any updates.
Let’s say we want to add Tests directory later to the config file... That would be not possible

OskarStark: Ok but in this case we would not get any updates.
Let’s say we want to add Tests directory later to the config file... That would be not possible

OskarStark: Ok but in this case we would not get any updates.
Let’s say we want to add Tests directory later to the config file... That would be not possible

OskarStark: What we can do is exclude the file via projects.yaml for specific bundles

Your case can be done by adding the file, run dev kit and deploy, remove the file from dev kit.

OskarStark: @core23 what do you think about it?

Another nice solution would be to extend our `Configuration` for the `projects.yaml` to support all necessary keys of the `phpstan.neon.dist` e.g.

Based on the BlockBundle:
```neon
parameters:
    level: 3

    paths:
        - src
# Uncomment this, when phpstan extensions are allowed
#        - tests

    excludes_analyse:
        - src/Test/BlockServiceTestCase.php
        - tests/bootstrap.php

    autoload_files:
        - vendor/autoload.php

    ignoreErrors:
        - '#Call to an undefined method Sonata\\Form\\Validator\\ErrorElement::assert.*\(\).#'
        - '#Call to an undefined method Symfony\\Component\\Config\\Definition\\Builder\\NodeDefinition::fixXmlConfig\(\).#'
```
->
```yaml
block-bundle:
  phpstan:
      level: 3
      paths:
          - src
      excludes_analyse:
          - src/Test/BlockServiceTestCase.php
          - tests/bootstrap.php
      autoload_files:
          - vendor/autoload.php
      ignoreErrors:
          - '#Call to an undefined method Sonata\\Form\\Validator\\ErrorElement::assert.*\(\).#'
          - '#Call to an undefined method Symfony\\Component\\Config\\Definition\\Builder\\NodeDefinition::fixXmlConfig\(\).#'
  branches:
    master:
      php: ['7.3', '7.4']
      target_php: '7.3'
      versions:
        symfony: ['4.3', '5.0']
    4.x:
      php: ['7.3', '7.4']
      target_php: '7.3'
      versions:
        symfony: ['4.3', '5.0']
```

In this case we can set `level: 1` and `paths: empty array` as default...
A bit more "complex" in the first step but much more flexible.

core23: > Another nice solution would be to extend our `Configuration` for the `projects.yaml` to support all necessary keys of the `phpstan.neon.dist` e.g.

I like the `project.yaml` solution, but we should not mix the baseline (ignored errors) and the configuration. The dev-kit shouldn't provide some default `ignore` entries, because not every project is a symfony lib that reports this errors.

OskarStark: but the ignored errors are configured per project...

core23: > but the ignored errors are configured per project...

IMHO that's intended

OskarStark: >  The dev-kit shouldn't provide some default ignore entries, because not every project is a symfony lib that reports this errors.

Sorry I don't understand your sentence here, can you please be more clear what we should do and what not? Thanks

core23: We should not do this in the `project.yaml` configuration:

```
      ignoreErrors:
          - '#Call to an undefined method Sonata\\Form\\Validator\\ErrorElement::assert.*\(\).#'
          - '#Call to an undefined method Symfony\\Component\\Config\\Definition\\Builder\\NodeDefinition::fixXmlConfig\(\).#'
```

Providing some default `ignoreErrors` is a bad idea, because then we should also set `reportUnmatchedIgnoredErrors: true`. In this case we will get an outdated ignore list over time. Some projects do not have these errors (e.g. php libs).

OskarStark: Ok but why is it in the the block bundle itself? If I configure this:
```yaml
 ignoreErrors:
          - '#Call to an undefined method Sonata\\Form\\Validator\\ErrorElement::assert.*\(\).#'
          - '#Call to an undefined method Symfony\\Component\\Config\\Definition\\Builder\\NodeDefinition::fixXmlConfig\(\).#'
```

for the block bundle, only the block bundle gets this configuration and not any other sonata lib. Or are you saying that it should be nowhere and we should remove it from the config in the block bundle too?

core23: Oh sorry, I didn't see that this is for the block bundle only. 

In any case, we should not configure the `ignoreErrors` in any case by the dev-kit, because this will become outdated very soon.

OskarStark: > Oh sorry, I didn't see that this is for the block bundle only.

OK

> In any case, we should not configure the ignoreErrors in any case by the dev-kit, because this will become outdated very soon.

But in this case the BlockBundle would fail, right?

franmomu: If the `phpstan.neon.dist` includes a file that does not exist, will it fail? (To include a `baseline.neon`).
(On my phone)

OskarStark: Thought about the same

franmomu: It fails if it doesn't exist 😞. But [there is `bootstrap` parameter](https://github.com/phpstan/phpstan#bootstrap-file) which could create the `phpstan-baseline.neon` if it doesn't exist.

jordisala1991:
@OskarStark is there any way to use your action and be able to  use phpstan extensions? IMO there are some useful extensions like symfony, and phpunit.

OskarStark:
Sure, they are already included in the action

OskarStark:
@jordisala1991 what du you think of merging it and add a config to projects.yaml if this should be a required check or not? 

This way we could fix issues project by project.

Also a config option for the used phpstan level could be added 🧐

jordisala1991:
IMO having this here is good, but the level option I dont know. I was thinking on adding config option on each project to enable phpstan and psalm. Then the file is created but the contents are based on those flags. The required check will only be added if the flag is true

jordisala1991:
but every other thing is handled per project for now. Because level could be based on branch too, see admin bundle, where we have config files inside dir, and on 3.x we can go beyond level 5 but on master we will get level 8

jordisala1991:
We should probably close this one, this is already implemented on dev-kit

Issue #660, opened 2020-04-10
?object $object = null VS object $object = null


VincentLanglet:
Can we make a decision for the `nullable_type_declaration_for_default_null_value` rule of phpcs-fixer ?

@sonata-project/contributors  Do you prefer 
```
function foo(?object $object = null)
```
Or
```
function foo(object $object = null)
```

I personally prefer the first object, this way changing the default value doesn't change the allowed types. But I you prefer the second one, I'll go with it !

core23:
👍 for the first one. So you can activly set a null value

greg0ire:
I prefer the first one too, it's more explicit.

phansys:
I tend to follow this convention:
* Nullable type declaration where the argument is nullable but explicitly required:
```php
function foo(?object $requiredObject)
```
* `null` as default value where the argument is nullable and optional:
```php
function foo(bool $requiredArg, object $optionalObject = null)
```
I'd avoid this form, since it will always make the argument optional:
```php
function foo(?object $requiredObject = null)
```

VincentLanglet:
> I'd avoid this form, since it will always make the argument optional:
> 
> ```
> function foo(?object $object = null)
> ```

I don't understand, because when you write
```
function foo(object $object = null)
```
the argument is also **always** optional.

IMO, the default value should not have impact on the allowed value. Changing
```
final function foo(?bool $bool = null)
```
To
```
final function foo(bool $bool = false)
```
would be a BC-break when changing 
```
final function foo(?bool $bool = null)
```
To
```
final function foo(?bool $bool = false)
```
is not always one.

Since there is 9 contributors, I'll wait for 5 vote in favor of one of the two forms.

phansys:
>I don't understand

Let's use a more realistic example:
```php
function isGranted(?object $object, bool $optionalArg = null)
{
    if (null === $object) {
        throw MyCustomDomainException();
    }

    if (null === $optionalArg) {
        // do something ...
    }
}
```
Based on this, we can make the following assumptions about the argument 1:
* Accepts an object or a null value;
* Is required;
* Calling the method without arguments will fail (`ArgumentCountError: Too few arguments to function isGranted()`).

If we force the codebase to always use the nullable type declaration (`?`) for every argument that uses `null` as default value, we'll lose the ability to distinguish between required and optional nullable arguments:
```php
function isGranted(?object $object = null, bool ?$optionalArg = null)
```

VincentLanglet:
[...]

PR #661, opened 2020-04-10
Add nullable_type_declaration_for_default_null_value rule

VincentLanglet:
Related to https://github.com/sonata-project/dev-kit/issues/660

Will be merged only If I get 5 approval in the issue, actually there is four.

By the way, what'll be the next step ? 
Won't the others projects build fail since they actually does not respect the rule ?

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -31,6 +31,7 @@ $rules = [
     'no_php4_constructor' => true,
     'no_useless_else' => true,
     'no_useless_return' => true,
+    'nullable_type_declaration_for_default_null_value' => ['use_nullable_type_declaration' => true],
     'ordered_class_elements' => true,
     'ordered_imports' => true,
     'phpdoc_order' => true,
--- a/project/.php_cs.dist.twig
+++ b/project/.php_cs.dist.twig
@@ -32,6 +32,7 @@ $rules = [
     'no_superfluous_phpdoc_tags' => ['allow_mixed' => true],
     'no_useless_else' => true,
     'no_useless_return' => true,
+    'nullable_type_declaration_for_default_null_value' => ['use_nullable_type_declaration' => true],
     'ordered_class_elements' => true,
     'ordered_imports' => true,
     'phpdoc_order' => true,


.php_cs.dist
```diff
@@ -31,6 +31,7 @@ $rules = [
     'no_php4_constructor' => true,
     'no_useless_else' => true,
     'no_useless_return' => true,
+    'nullable_type_declaration_for_default_null_value' => true,
```
phansys, line 34:
```suggestion
    'nullable_type_declaration_for_default_null_value' => ['use_nullable_type_declaration' => true],
```

greg0ire:
> actually

You mean "currently", actually is a _faux ami_ ;)

>  Won't the others projects build fail since they actually does not respect the rule ?

What we usually do is add commit to the dev-kit branch so that the build passes.

VincentLanglet:
> > actually
> 
> You mean "currently", actually is a _faux ami_ ;)

Sometimes I wonder if people don't bleed eyes when they read me 😅

.php_cs.dist
VincentLanglet: @phansys Is `'nullable_type_declaration_for_default_null_value' => true,` invalid or are you asking for this change to avoid BC-break if the default value change ?

phansys: The 2nd reason, plus I think we should be as verbose as possible here.

VincentLanglet:
> > Won't the others projects build fail since they actually does not respect the rule ?
> 
> What we usually do is add commit to the dev-kit branch so that the build passes.

So I merge this, a PR will be created to update the others projects and I fix the PR for every project ?

greg0ire:
> So I merge this, a PR will be created to update the others projects and I fix the PR for every project ?

Yes. In the case of CS fixes, I think FlintCI will open PRs targeting the dev-kit PR branches, so the fix will just be merging those PRs

OskarStark:
Yes, but currently this is just the config for the code of the dev kit itself

VincentLanglet:
> Yes, but currently this is just the config for the code of the dev kit itself

Indeed, I fixed my PR. Thanks @OskarStark !

VincentLanglet:
@greg0ire I'm not auhorized to merge this. ;)

greg0ire:
I wasn't expect you to merge your own PR :P Anyway, I'll run a deploy, thanks for this!

VincentLanglet:
> I wasn't expecting you to merge your own PR :P Anyway, I'll run a deploy, thanks for this!

Shouldn't I merge my own PR ?
I already merged some of them after two approvals, I understood this was ok 😶

PR #680, opened 2020-05-08
Ignore node_modules path

core23:
The `node_modules` path should not be commited.


OskarStark:
@greg0ire can you please merge and deploy dev-kit?

PR #701, opened 2020-05-25
Small improvements for `PULL_REQUEST_TEMPLATE.md`

phansys:


project/.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -28,16 +28,16 @@ Closes #{put_issue_number_here}
...
-    a release, you can delete the whole Changelog section.
+    a release, you can delete the whole "Changelog" section.
     (eg. update to docs, tests)
 -->
 
 <!-- REMOVE EMPTY SECTIONS -->
 ```markdown
 ### Added
-- Added some `Class::newMethod` to do great stuff
+- Added some `Class::newMethod()` to do great stuff.
```
core23, line 40: Do we need the brackets? If yes, should we add the method signature too?

phansys: It's not such a need. I'd say it's more a preference on my side. I consider easy to distinguish a method from a constant (or even a class property, regardless the dollar sign) if the method references always use the brackets. For the described purpose, the method signature should be not required.
By instance:
```php
Class::newMethod // could be a valid constant, regardless the case is not respecting the CS.
Class::newMethod() // is easily identified as a method.
```

core23: What about this:
```
- Added some `Class::newMethod` method to do great stuff. 
```
So you know there's a new method, but you do not need to pass the whole method signature

phansys: >For the described purpose, the method signature should be not required.

IMO, using just the brackets is more concise and enough to tell it's a method, the signature is not required.
I think that way is consistent on how PHP makes reference to functions or methods in its own messages.
By instance (https://3v4l.org/HsdNZ):
>Return value of Foo::bar() must be of the type int, string returned

But, if you prefer, we could indicate explicitly that it's a method instead.

greg0ire: @core23 @phansys please resolve this so that we can deploy

phansys: @core23, could we use the brackets?

PR #705, opened 2020-05-26
Add "logical_operators" to PHP CS Fixer rules

phansys:
This rule should cover cases like this: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/839#discussion_r259950505.


greg0ire:
Thanks @phansys !

PR #777, opened 2020-07-04
Fix twig-extension doc build

wbloszyk:
This change will allow generate documentation for twig-extension without build fail.


wbloszyk:
BTW 
Should the same be done for form-extensions too?

jordisala1991:
yes

jordisala1991:
Thank you @wbloszyk

PR #783, opened 2020-07-17
Revert support for 3.x BlockBundle

wbloszyk:
This branch should be suppport as long as Sonata 4 is not finish.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -114,6 +114,13 @@ block-bundle:
       target_php: '7.3'
       versions:
         symfony: ['4.4', '5.0']
+    3.x:
+      php: ['7.2', '7.3', '7.4']
+      target_php: '7.3'
+      versions:
+        symfony: ['4.4']
+        sonata_admin: ['3']
+        sonata_core: ['3']
 
 classification-bundle:
   branches:


jordisala1991:
Careful with this one, it means dev-kit will start trying to merge 3.x into 4.x

wbloszyk:
It is any possible to disable this auto-merging or we should upgrade dev-kit manualy for 3.x-block-bundle each time? @jordisala1991

VincentLanglet:
> Careful with this one, it means dev-kit will start trying to merge 3.x into 4.x

If a bugfix is made on 3.x, it should be backport in 4.x too. So maybe it's the right thing to do...

jordisala1991:
I guess so, but we are going to have an initial merge that will be huge. Not possible to disable them Atm

wbloszyk:
> > Careful with this one, it means dev-kit will start trying to merge 3.x into 4.x
> 
> If a bugfix is made on 3.x, it should be backport in 4.x too. So maybe it's the right thing to do...

When auto-merge fail then conflict exists and should be done by manual? If yes then will be awesome becouse we have the same minimum requirements.

> I guess so, but we are going to have an initial merge that will be huge. Not possible to disable them Atm

36 files to merge, 11 to delete. It is not so huge for me. I can do it in Monday.

greg0ire:
Let's merge this then.

PR #784, opened 2020-07-18
Add compatibility with SonataAdmin master branch

VincentLanglet:


config/projects.yml
```diff
@@ -214,11 +214,10 @@ doctrine-mongodb-admin-bundle:
 doctrine-orm-admin-bundle:
   branches:
     master:
-      php: ['7.2', '7.3', '7.4']
+      php: ['7.3', '7.4']
       target_php: '7.3'
       versions:
         symfony: ['4.4']
-        sonata_admin: ['3']
```
jordisala1991, line 221: maybe it is better to use 4 here

VincentLanglet: I don't know how this config works, but the version 4 doesn't exist.

VincentLanglet:
```
- php: '7.3'
      env: SONATA_ADMIN=4.*
```
will fail

jordisala1991: This works by adding an extra build for 4.x-dev of sonata admin. I guess now it is not required , but when you release 4, it is better to also test with unreleased 4.x (to ensure the latest Pr does not break compatibility)

VincentLanglet: Yes, but it does 2 things:
```
{% for package_name,package_versions in versions %}
{% for version in package_versions %}
    - php: '{{ target_php|default(php|last) }}'
      env: {{ package_name|upper }}={{ version }}.*
{% endfor %}
{% if package_versions|length > 0 %}
    - php: '{{ target_php|default(php|last) }}'
      env: {{ package_name|upper }}='dev-master as {{ package_versions|last }}.x-dev'
{% endif %}
{% endfor %}
```
Will add 
```
- php: '7.3'
   env: SONATA_ADMIN=4.*
```

And 
```
{% for package_name,package_versions in versions %}
{% if package_versions|length > 0 %}
    - env: {{ package_name|upper }}='dev-master as {{ package_versions|last }}.x-dev'
{% endif %}
{% endfor %}
```
Will add an extra build for 4.x-dev.

If we want the second, the first will be added and will obviously fail.
So we need to wait for the 4.x release in order to re-add this.

VincentLanglet:
@OskarStark @greg0ire Can you merge this ? :)

config/projects.yml
core23, line 221: Should we add `dev-master` to test against the master branch?

VincentLanglet: The composer.json is already using dev-master, this would be a duplicate build then.

greg0ire:
Can you please improve your commit message to explain why and how this makes the orm bundle job suite compatible with the sonata admin master branch? Are both changes strictly related to that? It's not super obvious to me…

VincentLanglet:
@greg0ire Is it better ?

greg0ire:
It is, but why did you make 2 commits? One does not make sense without the other, right? Commit messages can have a body in case you don't know.

VincentLanglet:
> It is, but why did you make 2 commits? One does not make sense without the other, right? Commit messages can have a body in case you don't know.

I thought you wanted two commit. Fixed.

greg0ire:
Ok but now the commit messages is unhelpful… I think I understood what this is about, and changed it. Please tell me if that is indeed what you mean to do, and if yes, why we stop running tests with v3.

VincentLanglet:
> Ok but now the commit messages is unhelpful… I think I understood what this is about, and changed it. Please tell me if that is indeed what you mean to do,

It is.

> and if yes, why we stop running tests with v3.

I just to on master branch.
SonataDoctrineORMAdmin v3 will still run test with SonataAdmin v3

But I'm making a PR to use SonataDoctrineORMAdmin dev-master and SonataAdmin dev-master https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1073

If I keep running test with SonataAdmin v3 (on master branch), it will try to use SonataDoctrineORMAdmin dev-master and SonataAdmin v3. This won't work.

greg0ire:
Thanks @VincentLanglet !

PR #792, opened 2020-07-25
Always use twig template to generate file and add a conditional abandoned topic on readme

VincentLanglet:
Related to https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/601

I wanted to add an `Abandoned` topic on the readme.

If I understand correctly, it works with `DispatchCommand::renderFile`.

I discover there was a way to handle `.twig` template, and another way for others files with `str_replace`. I feel like it's more consistent to always use `twig` for template which need variables/conditions. (And it avoid to pass variables to both twig and str_replace).
Moreover, in my case, I added a variable `is_abandoned` to display or not some text on the Readme.


project/README.md.twig
```diff
@@ -15,6 +15,14 @@ Branch | Travis | Coveralls | Scrutinizer |
 {{ stable_branch }}   | [![Build Status][travis_stable_badge]][travis_stable_link]     | [![Coverage Status][coveralls_stable_badge]][coveralls_stable_link]     | [![Scrutinizer Status][scrutinizer_stable_badge]][scrutinizer_stable_link] |
 {{ unstable_branch }} | [![Build Status][travis_unstable_badge]][travis_unstable_link] | [![Coverage Status][coveralls_unstable_badge]][coveralls_unstable_link] | [![Scrutinizer Status][scrutinizer_unstable_badge]][scrutinizer_unstable_link] |
 
+{% if is_abandoned %}
+## WARNING: This bundle is abandoned
```
greg0ire, line 19: It could also be a library, couldn't it?

project/README.md.twig
```diff
@@ -15,6 +15,14 @@ Branch | Travis | Coveralls | Scrutinizer |
...
+
+Since maintainers have lost interest in this project, there is no active support on it.
```
greg0ire, line 21: If all maintainers have lost interest in a project, but there still are a lot of contributions, would we drop support for it?

jordisala1991: I think it would be convenient to leave a call for those contributors to step up and become mantainers

VincentLanglet: Right, I was trying to make a generic sentence. Do you have any suggestion ?
Should I only write `There is no active support`, then ?

greg0ire: > I think it would be convenient to leave a call for those contributors to step up and become mantainers

That is what is done below, isn't it?

project/README.md.twig
core23, line 19:
```suggestion
## WARNING: This project is abandoned
```

project/README.md.twig
```diff
@@ -15,6 +15,14 @@ Branch | Travis | Coveralls | Scrutinizer |
...
+
+Feel free to ask if you want to help to keep this project up to date.
+
+{% endif %}
 ## Documentation
 
 Check out the documentation on the [official website](https://sonata-project.org/bundles/{{ website_path }}).
```
core23, line 28: Not related to this PR, but AFAIK we don't have a valid documentation for all projects

project/README.md.twig
VincentLanglet: Should I change the sentence then ? And how ?

I thought again, and I think it's ok.
- I ask for new maintainer in the next sentence.
- If we drop support on a project, we have indeed lost interest in it. I don't say that we'll abandon every project we lost interest in.

project/README.md.twig
VincentLanglet: Indeed, for example: https://github.com/sonata-project/form-extensions

project/README.md.twig
greg0ire: You don't explicitely say it but it seems to be the implication.
```suggestion
There is no active support on it.
```

VincentLanglet: Ok, I updated the message then.

project/README.md.twig
phansys: What about "package"? Some small packages maybe aren't a project itself (bridges, bundle implementations of a library, etc).

VincentLanglet: This thing 😄

greg0ire: Yeah let's go with repository, because we do have projects that are not packages (the sandbox)

phansys: From the [Composer perspective](https://getcomposer.org/doc/04-schema.md#type), all the repositories are considered packages (then, there are package types, like libraries and so).

greg0ire: Ok I'm fine with both

VincentLanglet: I feel like more logic the fact that
- The repository will say that the repository is abandoned
- The package will say that the package is abandoned

This will also say that we will not necessarily be as quick to respond as on another repository.

src/Command/DispatchCommand.php
```diff
@@ -13,7 +13,6 @@
 
 namespace App\Command;
 
-use Doctrine\Common\Inflector\Inflector;
```
jordisala1991, line 16: Is this used somewhere else, can this dependency be dropped on composer.json?

VincentLanglet: Done

symfony.lock
```diff
@@ -69,7 +69,7 @@
         ]
     },
     "php": {
-        "version": "7.3"
+        "version": "7.4"
```
jordisala1991, line 72: not sure about this change

composer.lock
```diff
@@ -262,21 +261,7 @@
                 "uppercase",
                 "words"
             ],
-            "funding": [
```
jordisala1991, line 265: I think your composer version is a bit old

VincentLanglet: Composer version 1.10.9 2020-07-16 12:57:00

symfony.lock
VincentLanglet: Automatically made when I ran `composer remove doctrine/inflector`.

Lock files are not committed in other project.
Since it depends on the environment, I'm not sure what to do...

jordisala1991: On bundles it should not be commited, but on projects that get deployed it is normal. IMO you should revert this change

composer.lock
jordisala1991: Can you try with composer 2?

VincentLanglet: Funny, removing the vendors fixed my issue.

VincentLanglet:
I'll need to rebase after https://github.com/sonata-project/dev-kit/pull/807

VincentLanglet:
Rebased and RTM if you'll have time to take a look @OskarStark

OskarStark:
Can you please rebase @VincentLanglet

VincentLanglet:
Sure, but I'll wait one day or two, since @jordisala1991 are doing a lot of changes :)

VincentLanglet:
@OskarStark It's rebased.

OskarStark:
Please rebase again, I deleted some files. Also why is composer.lock modified? 🤔

VincentLanglet:
Seems like it's adding what was removed here: https://github.com/sonata-project/dev-kit/pull/821/files#diff-1da2c7edc898c70e5a79a9997c98cecc

OskarStark:
but also the files:
templates/project/.github/DELETE_ISSUE_TEMPLATE.md

are now removed in master, no need to rename them anymore

VincentLanglet:
Done @OskarStark

OskarStark:
I will deploy and see what happens 👍  Wait a minute....

PR #793, opened 2020-07-26
Remove builds with unstable code of SonataCoreBundle

jordisala1991:
We don't need to add more builds for the 3.x and master branches
of SonataCoreBundle, we are already testing on the --prefer-lowest
with the latest tag of SonataCoreBundle, and since this bundle
is deprecated, that should be enough.

Bonus: Faster builds

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -69,7 +69,6 @@ admin-bundle:
       target_php: '7.3'
       versions:
         symfony: ['4.4']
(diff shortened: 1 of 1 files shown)


VincentLanglet:
I'm wondering.

Can't we remove the same way the lines `sonata_block: ['3']` or  `sonata_admin: ['3']` ?

PR #799, opened 2020-08-03
Use constructor injection instead of creating new instance

VincentLanglet:
Related to https://github.com/sonata-project/dev-kit/pull/797#issuecomment-667840068

The freshly created twig extension is not know by the TwigEnvironment used by this command.
Instead of creating new instance of service in the initialize method, I use dependency injection instead.

This may fix the issue, but I don't know how to try...

PR #800, opened 2020-08-04
Fix injection of git wrapper

VincentLanglet:
Fix
https://github.com/sonata-project/dev-kit/pull/799#issuecomment-668169593

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -27,3 +27,4 @@ services:
 
     # add more service definitions when explicit configuration is needed
     # please note that last definitions always *replace* previous ones
+    GitWrapper\GitWrapper:


core23:
Can't you add a new `GitWrapper` service, so we could use autowiring?

VincentLanglet:
Like this @core23 ?

OskarStark:
Can you please deploy @greg0ire

OskarStark:
I deployed!

OskarStark:
Hmm...

Screenshot 2020-08-04 18 16 04

VincentLanglet:
Before it was 
```
$this->twig = new Environment(new FilesystemLoader(__DIR__.'/../..'));
```

now, with the injection, seems like it's only looking inside `/templates`.

What would be the best solution @sonata-project/contributors ?
- moving `project` folder inside the `template` folder
- changing `$this->twig->render($localPath, ...)` to `$this->twig->render('../'.$localPath, ...)` 
- something else ?

I think the first solution is not a bad one, after all, all the file inside the `project` dir are what we can call templates.

jordisala1991:
https://symfony.com/doc/current/reference/configuration/twig.html#paths

Configure another twig path

VincentLanglet:
> https://symfony.com/doc/current/reference/configuration/twig.html#paths
> 
> Configure another twig path

`project` dir is in the root dir. I'm not sure that adding the root dir of the project as a twig path is nice.

Plus, the signature of the method use to render a file is
```
private function renderFile(Package $package, $repositoryName, $localPath, $distPath, array $projectConfig, $branchName): void
```

Currently we're calling this with `$localPath = 'project'`.
Is does not seems stupid to me to require that `$localPath` should refer to a path inside the `templates` dir. (Or it can be inside another dir if you prefer and we add it as a twig path, but IMO it shouldn't be directly in the rootDir).

VincentLanglet:
> > https://symfony.com/doc/current/reference/configuration/twig.html#paths
> > Configure another twig path
> 
> `project` dir is in the root dir. I'm not sure that adding the root dir of the project as a twig path is nice.
> 
> Plus, the signature of the method use to render a file is
> 
> ```
> private function renderFile(Package $package, $repositoryName, $localPath, $distPath, array $projectConfig, $branchName): void
> ```
> 
> Currently we're calling this with `$localPath = 'project'`.
> Is does not seems stupid to me to require that `$localPath` should refer to a path inside the `templates` dir. (Or it can be inside another dir if you prefer and we add it as a twig path, but IMO it shouldn't be directly in the rootDir).

Hi @sonata-project/contributors. I'll want to make the PR to fix this.
But I would prefer to have your opinion about what is the best solution.

OskarStark:
Would move the folder in templates folder 👍🏻

PR #801, opened 2020-08-06
Move twig template inside template dir

VincentLanglet:
I'm trying to fix https://github.com/sonata-project/dev-kit/pull/800#issuecomment-668691055


.flintci.yml
```diff
@@ -1 +0,0 @@
-project/.flintci.yml
```
jordisala1991, line 1: I think we could maintain those symlinks but with the new location right?

VincentLanglet: What about using the same process than other Sonata bundle and use the DispatchCommand to update these files instead ?

VincentLanglet: I can still reuse a symbolic link (just need to look at the command online ^^'), but this won't work the day this file will be change to `.flintci.yml.twig`

jordisala1991: > What about using the same process than other Sonata bundle and use the DispatchCommand to update these files instead ?

Well we could do it, but it will be a bit problematic because the dispatcher dispatch a lot of files and we only use those 2 here.

I would say to maintain the symlinks for now (to have dev-kit) working again and finish the orm master branch with the admin) and think on how to improve this later.

And, if we have just 2 files, when they become twig, we can just duplicate them and not worry about adding more code.

VincentLanglet: Done.
I did the same for `.php_cs.dist` and `.editorconfig`.

VincentLanglet:
I don't find your comment @OskarStark

OskarStark:
Oh sorry

.editorconfig
```diff
@@ -1,14 +0,0 @@
-root = true
```
OskarStark, line 1: Are you sure this works to use a filepath, for PHPStorm etc. ? 🤔

I would not touch the files from dev kit itself in this PR and would Not couple the project files with the dev kit app anyway. In this case I would prefer duplication

VincentLanglet: Do you want duplication for all of config file (phpcs, yamllint, flintci, ...) or just for editorconfig ?

OskarStark: I just want to keep it untouched we’re we did t use paths like now. Anyway are you sure this works for php cs fixer? 🤔

VincentLanglet: It work like this for editorconfig.

It wasn't for phpcs. I made a fix.

VincentLanglet:
I also add a `DO NOT EDIT THIS FILE` header for multiple templates files.
1) For consistency
2) To avoid edition in other repository

I let without header
- the license
- the .github/ files, since it's used for PR/issue.

templates/project/.php_cs.dist
```diff
@@ -42,16 +42,12 @@ $rules = [
     '@PHP56Migration:risky' => true,
     '@PHPUnit57Migration:risky' => true,
     'single_line_throw' => false,
-{% if php|first in ['7.0', '7.1', '7.2', '7.3'] %}
```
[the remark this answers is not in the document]
VincentLanglet: This check was "If php >= 7.0". The next one was "if php >= 7.1".

You can see here: https://github.com/sonata-project/dev-kit/blob/master/config/projects.yml
All our lib start with at least 7.2.

templates/project/docs/conf.py.twig
```diff
@@ -1,3 +1,7 @@
+# DO NOT EDIT THIS FILE!
+#
+# It's auto-generated by sonata-project/dev-kit package.
+
 # -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
VincentLanglet: I don't know. So I change the order.

OskarStark:
Ready to merge from my side 👍🏻 Great work, thank you

VincentLanglet:
Can you make a deploy @OskarStark ? :)

OskarStark:
Currently not, on vacation.

I am currently on a phone 📱 

Cc @greg0ire

greg0ire:
Done :)

PR #804, opened 2020-08-09
Leverage the GitHub's template chooser

phansys:
See "[Configuring the template chooser](https://docs.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser)".

PR #805, opened 2020-08-10
Remove files prefixed with "DELETE_"

phansys:
Follow up of #804.


OskarStark:
We need to revert this one, deploy and run dev kit and then do this PR again.... thanks

PR #810, opened 2020-08-13
Fix dispatch command

VincentLanglet:
The dispatch command is currently failing because of an error ligne 527.

I think it's because the `fileperms` method does not find the file since I move it inside the template dir.
So I replace every usage of the relative path by the full path.

Moreover, I find a check
```
\in_array(substr($localPath, 8), $projectConfig['excluded_files'], true)) {
```
This only works if the dir is called `project` (7 letters, 8 with the `/` symbol).

So I preferred to
- create a constant for the `project` dir.
- add a check at the start of the `renderFile` that the local path is starting by `project/`.

### DIFF
--- a/src/Command/DispatchCommand.php
+++ b/src/Command/DispatchCommand.php
@@ -494,7 +494,7 @@ final class DispatchCommand extends AbstractNeedApplyCommand
         }
 
         $branchConfig = $projectConfig['branches'][$branchName];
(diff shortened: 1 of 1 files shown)


OskarStark:
Looks good.

@greg0ire cab you please deploy?

greg0ire:
Done!

OskarStark:
Thank you

PR #811, opened 2020-08-14
Add exceptions to debug

VincentLanglet:
A lot of php methods are returning false.

I think it's better to throw an exception instead of failing after because some functions expect int and get false.


greg0ire:
Thanks, deployed!

PR #821, opened 2020-08-15
Use monolog handler for Slack over separate client

greg0ire:
This results in fewer dependencies and environment variables.

### DIFF
--- a/config/packages/prod/monolog.yaml
+++ b/config/packages/prod/monolog.yaml
@@ -32,3 +32,11 @@ monolog:
             bot_name: '@@dev-kit-logger'
             level: debug
             include_extra: true
+        slack:
+            type: slack
+            token: '%env(SLACK_TOKEN)%'
+            channel: 'ci'
+            bot_name: '@@dev-kit-logger'
+            level: debug
+            include_extra: true
+            channels: ['automerge']
--- a/config/packages/prod/monolog.yaml
+++ b/config/packages/prod/monolog.yaml
@@ -30,6 +30,5 @@ monolog:
             token: '%env(SLACK_TOKEN)%'
             channel: 'dev-kit-logs'
             bot_name: '@@dev-kit-logger'
-            icon_emoji: ':ghost:' # 👻
             level: debug
             include_extra: true
--- a/.env.dist
+++ b/.env.dist
@@ -1,6 +1,5 @@
 GITHUB_OAUTH_TOKEN=token_to_be_changed
 DEK_KIT_TOKEN=token_to_be_changed
-SLACK_HOOK_URL=hook_to_be_changed
 SLACK_TOKEN=token_to_be_changed
 
 ###> symfony/framework-bundle ###
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,6 @@
         "doctrine/annotations": "^1.8",
         "knplabs/github-api": "2.5",
         "knplabs/packagist-api": "^1.3",
-        "maknz/slack": "^1.7",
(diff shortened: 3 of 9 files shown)


config/packages/prod/monolog.yaml
```diff
@@ -33,3 +33,12 @@ monolog:
             icon_emoji: ':ghost:' # 👻
             level: debug
             include_extra: true
+        slack:
+            type: slack
+            token: '%env(SLACK_TOKEN)%'
+            channel: 'dev-kit-logs'
+            bot_name: '@@dev-kit-logger'
+            icon_emoji:  ":hankey:" # 💩
```
FlintCIBot, line 41: - yamllint: too many spaces after colon (colons)

config/packages/prod/monolog.yaml
```diff
@@ -30,6 +30,13 @@ monolog:
             token: '%env(SLACK_TOKEN)%'
             channel: 'dev-kit-logs'
             bot_name: '@@dev-kit-logger'
-            icon_emoji: ':ghost:' # 👻
             level: debug
             include_extra: true
+        slack:
+            type: slack
+            token: '%env(SLACK_TOKEN)%'
+            channel: 'dev-kit-logs'
```
VincentLanglet, line 38: I have trouble to understand the config.

Why is it again the `dev-kit-logs` channel and the dev-kit-logger ?
But then there is a `channels` key ?

greg0ire: That's a Slack channel. I think we could archive `#ci` and put the dev-kit logs in dev-kit-logger

VincentLanglet: When I read this, I understood that the `#ci` channel was replaced by a `#automerge` channel.

Now I understand that with this config:
- The Slack_errors key will listen all the logs and post them in the `dev-kit-logs` slack channel.
- The Slack key will listen all the logs from the `automerge` monolog channel and post them in the `dev-kit-logs` slack channel.

I'm surely have a bad understanding because it would lead to duplicate logs...

greg0ire: > The slack_errors key will listen all the logs and post them in the dev-kit-logs slack channel.

`slack_errors` is nested inside other handlers, the top-level one being of type `fingers_crossed`, and set with to be triggered with `error`. It will be only be triggered if there is an error or above beforehand.

The only case where you could have double posting is when we log an error or above in the `automerge` monolog channel.

VincentLanglet: Ok. I see.

So now the only question I have is: Should we split the slack log by context (so having a dev-kit-logs for every errors and keeping/renaming the ci channel for the automerge warning) or should we have everything on dev-kit-logs like you did ?

`ci` channel was public, which could allow someone which is not in the `contributor` team to help us.

greg0ire: :thinking: good point… I can change to `ci`, and then we can say we should have one slack channel per monolog channel, because calls to `this->logger->whatever` are voluntary while `slack_errors` is meant for things we didn't plan for like exceptions.

VincentLanglet: We can change the strategy later if we decide too. But I think keeping the two channel is better ATM.
Plus it's BC 😁

VincentLanglet:
@greg0ire You amend the emoji commit.

Not sure someone expect to find a channel change in the emoji commit

greg0ire:
Oh right I meant the other way around, obviously. I changed the order of the commits, it's safer and better anyway.

PR #827, opened 2020-08-23
remove scrutinizer due it's deprecation by it's authors

vladyslavstartsev:
During implementation of https://github.com/sonata-project/SonataAdminBundle/pull/6288 I've found out that I broke `scrutinizer-ci`. When investigating the reason why I've found that, internally it (the scrutinizer-ci) uses outdated version of `composer-plugin-api` (the error is similar to https://github.com/composer/composer/issues/2324).

There are a list of reasons for removing it, see it here https://github.com/sonata-project/SonataAdminBundle/pull/6288#issuecomment-678785667

This PR only removes it (the scrutinizer-ci) for new repositories


core23 (approved):

OskarStark:
> We also need to remove the we hook by the command. It needs some manual work, I can do that

I am removing the hook in #836

OskarStark:
All hooks are removed, so lets merge this one 👍 

Thank you @vladyslavstartsev

PR #836, opened 2020-08-27
Enhancement: Add ability to delete hooks

OskarStark:


src/Command/DispatchCommand.php
```diff
@@ -32,6 +32,13 @@ final class DispatchCommand extends AbstractNeedApplyCommand
     private const LABEL_NOTHING_CHANGED = 'Nothing to be changed.';
     private const FILES_DIR = 'project';
 
+    /** @var string[] */
+    private const HOOK_URLS_TO_BE_DELETED = [
+        'https://api.codacy.com',
```
OskarStark, line 37: I tested this with this hook and `admin-bundle` and it removes it like expected 👍

src/Command/DispatchCommand.php
```diff
@@ -32,6 +32,13 @@ final class DispatchCommand extends AbstractNeedApplyCommand
...
+        'https://www.flowdock.com',
+        'http://scrutinizer-ci.com',
```
OskarStark, line 39: @sonata-project/contributors are you fine with deleting these hooks?

src/Command/DispatchCommand.php
```diff
@@ -265,9 +285,38 @@ private function updateDevKitHook(Package $package): void
...
+
+                    if ($this->apply) {
+                        $this->githubClient->repo()->hooks()->remove(static::GITHUB_GROUP, $repositoryName, $hook['id']);
+
+                        $this->io->success(sprintf(
+                            'Hook "%s" deleted.',
+                            $currentHookUrl
+                        ));
+
+                        die();
```
greg0ire, line 309: Are you sure about this? :sweat_smile:

OskarStark: 😆  nope

OskarStark: Fixed

OskarStark:
Strange, which version of PHP-CS-Fixer is used by FlintCI? FlintCi is fine, but the GithubAction is complaining about the code style.

cc @soullivaneuh

greg0ire (dismissed):

src/Command/DispatchCommand.php
```diff
@@ -169,7 +183,13 @@ private function updateLabels(Package $package): void
         $configuredLabels = $this->configs['labels'];
         $missingLabels = $configuredLabels;
 
-        $headers = ['Name', 'Actual color', 'Needed Color', 'State'];
+        $headers = [
+            'Name',
+            'Actual color',
+            'Needed Color',
+            'State'
```
greg0ire, line 190:
```suggestion
            'State',
```

OskarStark: Yes, I am talking to @soullivaneuh why the GithubAction fails, but FlintCi is happy, so lets keep it as an example until we found out why! 👍

jordisala1991: IMO with all the checks we have now, we could consider removing flintCI.

jordisala1991 (approved):

src/Command/DispatchCommand.php
```diff
@@ -32,6 +32,13 @@ final class DispatchCommand extends AbstractNeedApplyCommand
     private const LABEL_NOTHING_CHANGED = 'Nothing to be changed.';
     private const FILES_DIR = 'project';
 
+    /** @var string[] */
+    private const HOOK_URLS_TO_BE_DELETED = [
+        'https://api.codacy.com',
+        'https://www.flowdock.com',
```
core23, line 38: What is flowdock? Never saw this on our projects

OskarStark: It was only on one project and deactivated...

core23 (approved):

src/Command/DispatchCommand.php
OskarStark: FlintCI creates PullRequests which are merged automatically, so we need to keep it until we fix, and commit the fixed files using a GithubAction. Lets postpone this a bit

VincentLanglet: > FlintCI creates PullRequests which are merged automatically, so we need to keep it until we fix, and commit the fixed files using a GithubAction. Lets postpone this a bit

I may be wrong but the 3.x or master branch are not supposed to have lint-failures since we only merge PR with a green build. And if someone open a PR with a lint-failure, the build will be red and we'll ask for fixing it before merging it.

Or maybe it's only useful only when we're adding new rules to linters ?

PR #850, opened 2020-08-28
Improve Github Actions

jordisala1991:
Now they look similar to other sonata-project bundles


OskarStark:
Do we really want to have PHP 8 test?

jordisala1991:
IMO it is nice to have it, one day we would like to switch, and having test with it helps the migration (even if it does not work now, probably because some incompatibility with SF)

config/packages/github_api.yaml
```diff
@@ -3,7 +3,9 @@ services:
         arguments:
             - '@Github\HttpClient\Builder'
         calls:
-            - ['authenticate', ['%env(GITHUB_OAUTH_TOKEN)%', null, !php/const \Github\Client::AUTH_HTTP_TOKEN]]
+            # use: !php/const \Github\Client::AUTH_HTTP_TOKEN instead of: 'http_token'
+            # when the yaml parser can deal with custom tags
+            - ['authenticate', ['%env(GITHUB_OAUTH_TOKEN)%', null, 'http_token']]
```
jordisala1991, line 8: Had to do this change because of the yaml-parser, it is bad that it does not support custom tags

OskarStark:
`Test / PHP 7.4 (pull_request)` should be required

jordisala1991:
This has to be configured on the settings of this repository, I dont have enough rights to do it

PR #854, opened 2020-08-29
Remove unneeded target php version

jordisala1991:
Now all the variants execute with php 7.4

It was added when php 7.4 was on snapshot: https://github.com/sonata-project/dev-kit/pull/515

I didn't remove the config option because it might be useful in some cases, but when everything gets stable, we might remove it too. (on another PR)

### DIFF
--- a/src/Config/ProjectsConfiguration.php
+++ b/src/Config/ProjectsConfiguration.php
@@ -75,7 +75,7 @@ class ProjectsConfiguration implements ConfigurationInterface
         $builder = new TreeBuilder('versions');
         $node = $builder->getRootNode();
 
-        $childrenNode = $node->children();
+        $childrenNode = $node->addDefaultsIfNotSet()->children();
 
         foreach ($this->devKitConfigs['packages'] as $key => $name) {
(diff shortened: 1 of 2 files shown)


src/Config/ProjectsConfiguration.php
```diff
@@ -75,7 +75,7 @@ private function addVersionsNode()
         $builder = new TreeBuilder('versions');
         $node = $builder->getRootNode();
 
-        $childrenNode = $node->children();
+        $childrenNode = $node->addDefaultsIfNotSet()->children();
```
jordisala1991, line 78: Also fixed a bug here, where if the package does not have versions (AKA test this package with symfony 4.4 and 5.1 for example) it gets no `version` key and it breaks on my env atleast, maybe on prod it does not, but still, it should be considered as a bug

greg0ire (approved):

OskarStark (approved):

OskarStark:
Please rebase

jordisala1991:
Done

greg0ire (approved):

PR #910, opened 2020-09-04
Enhancement: Use Payload value object instead of array

OskarStark:


src/Github/GithubHookProcessor.php
```diff
@@ -32,47 +32,43 @@ public function __construct(GithubClient $client)
...
-    public function processPendingAuthorLabel(Event $event, array $payload): void
+    public function processPendingAuthorLabel(Payload $payload): void
     {
-        if (!\in_array($payload['action'], ['created', 'synchronize'], true)) {
+        if (!\in_array($payload->action(), ['created', 'synchronize'], true)) {
             return;
         }
 
-        $issueKey = 'issue_comment' === $event->toString() ? 'issue' : 'pull_request';
+        if ($payload->commentAuthorId() !== $payload->issueAuthorId()) {
```
greg0ire, line 41: Let's introduce a method that does just that instead of exposing these properties?

src/Github/Domain/Value/Webhook/Payload.php
```diff
@@ -0,0 +1,123 @@
...
+        return self::fromArray(
+            json_decode($payload, true),
+            $event
+        );
+    }
+
+    public function event(): Event
+    {
+        return $this->event;
+    }
```
greg0ire, line 97: This one seems to be used in tests only

src/Github/GithubHookProcessor.php
OskarStark: Will do, but will change it step by step

src/Github/Domain/Value/Webhook/Payload.php
OskarStark: The method yes, the property not 🙃

src/Github/GithubHookProcessor.php
OskarStark: commentAuthorEqualsIssueAuthor() ?

Other naming proposals @greg0ire ?

franmomu: `isTheCommentFromTheAuthor`?

OskarStark: Love it, thanks

greg0ire: `isTheCommentFromTheAuthor`

Exactly the kind of high-level naming I was hoping for :)
That will make the code easier to understand 👍

PR #920, opened 2020-09-06
Automatically update some github metadata

VincentLanglet:
Related to https://github.com/sonata-project/twig-extensions/pull/108

There is multiple things to do before merging this PR:
- I use the package_description for the description of the repository, which means we may need to update the package description in the composer.json of our bundle (which is not a bad thing).
- This automatically add `[abandonned]` to the description, but there is no such things for `[deprecated]` since we remove/removed them from the dev-kit project list. Do we add deprecated project ?
- I didn't add the `topics` argument. What do you think about using the `keywords` from the composer.json ? We'll need to update the `composer.json` of course.

### DIFF
--- a/src/Command/Dispatcher/DispatchSettingsCommand.php
+++ b/src/Command/Dispatcher/DispatchSettingsCommand.php
@@ -18,6 +18,7 @@ use App\Config\Projects;
 use App\Domain\Value\Project;
 use Github\Client as GithubClient;
 use Github\Exception\ExceptionInterface;
+use Packagist\Api\Result\Package;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
@@ -73,6 +74,7 @@ final class DispatchSettingsCommand extends AbstractNeedApplyCommand
     {
         $repository = $project->repository();
         $branchNames = $project->branchNames();
+        $latestVersion = $this->getLatestPackageVersion($project->package());
 
         $repositoryInfo = $this->github->repo()->show(
             $repository->vendor(),
@@ -80,7 +82,11 @@ final class DispatchSettingsCommand extends AbstractNeedApplyCommand
         );
 
         $infoToUpdate = [
-            'homepage' => 'https://sonata-project.org',
+            'homepage' => $latestVersion->getHomepage() ?: 'https://sonata-project.org',
+            'description' => $latestVersion->isAbandoned()
+                ? '[Abandonned] '.$latestVersion->getDescription()
+                : $latestVersion->getDescription(),
+            'topics' => $latestVersion->getKeywords(),
             'has_issues' => true,
             'has_projects' => true,
             'has_wiki' => false,
@@ -110,5 +116,21 @@ final class DispatchSettingsCommand extends AbstractNeedApplyCommand
         } else {
             $this->io->comment(static::LABEL_NOTHING_CHANGED);
         }
+    }
+
+    /**
+     * Return the latest packagist version.
+     */
+    private function getLatestPackageVersion(Package $package): Package\Version
+    {
+        $versions = $package->getVersions();
+        $lastVersion = reset($versions);
+
+        if (false === $lastVersion) {
+            // This package was never released, we create a fake empty version
+            return new Package\Version();
+        }
+
+        return $lastVersion;
     }
 }


templates/project/.github/settings.yml.twig
```diff
@@ -0,0 +1,4 @@
+repository:
```
core23, line 1: I thought you would extend our dev-kit to do the stuff probot is doing.

Adding a config for an other config tool does not make any sense to me.

templates/project/.github/settings.yml.twig
```diff
@@ -0,0 +1,4 @@
...
+    name: {{ repository_name }}
+    description: {% if is_abandoned|default(false) %}[Abandonned] {% endif %}{{ package_description }}
+    homepage: https:/sonata-project.org
```
core23, line 4:
```suggestion
    homepage: https://sonata-project.org
```

core23 (changes requested):

templates/project/.github/settings.yml.twig
VincentLanglet: It's the same strategy than github actions, linters, ...

For instance, we didn't extends our dev-kit to do the stuff that github actions are doing.
We did extend our dev-kit to have a only one place where we add the github actions and it's dispatched in every projects.

Instead of opening 20 PR, you'll have just to open one on Webkit, and the dispatch will do the job for you.

Plus it allows to synchronize the packagist description/keywords and the github description/topics like I proposed.
It can also automatically add `[Abandonned]` as soon as you update the composer.json.
We could also synchronize the packagist homepage and the github homepage.

VincentLanglet:
It seems like we don't really need probot.
I think the `updateRepository` method is enough.

@OskarStark How can we try with SonataAdminBundle ?

src/Command/DispatchCommand.php
```diff
@@ -141,12 +141,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $repositoryName = $this->getRepositoryName($package);
+        $lastVersion = $this->getLastPackageVersion($package);
+
         $branches = array_keys($projectConfig['branches']);
         $this->io->section('Repository');
 
         $repositoryInfo = $this->githubClient->repo()->show(static::GITHUB_GROUP, $repositoryName);
         $infoToUpdate = [
-            'homepage' => 'https://sonata-project.org/',
+            'homepage' => $lastVersion->getHomepage() ?: 'https://sonata-project.org/',
```
core23, line 151: IMHO we could use a static value here unless we have different URLs

VincentLanglet: We could use a more specific value: https://github.com/sonata-project/SonataAdminBundle/blob/3.x/composer.json#L11

core23: Can you move this to the `projects.yaml` config?

core23 (changes requested):

src/Command/DispatchCommand.php
VincentLanglet: Yes I can, but why ?

Shouldn't we try to synchronize the composer.json and the github page ?

jordisala1991 (approved):

src/Command/DispatchCommand.php
```diff
@@ -141,12 +141,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            'description' => $lastVersion->isAbandoned()
+                ? '[Abandonned] '.$lastVersion->getDescription()
+                : $lastVersion->getDescription(),
```
jordisala1991, line 154: We could probably improve the composer.json tags and description, but this is better than before IMO.

jordisala1991 (approved):

src/Command/DispatchCommand.php
VincentLanglet: I would like to try to on admin bundle, @OskarStark may help me for that.

If it works as expected, I'll make PR to improve composer.json for every package.
And we'll need a new release for all of them.

OskarStark: Will help but I am on holiday again 🤪

Back at next Monday 🥳

core23 (approved):

core23:
We should remove the probot app after merging this.

src/Command/AbstractCommand.php
```diff
@@ -106,4 +106,19 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
 
         return str_replace('.git', '', end($repositoryArray));
     }
+
+    /**
+     * Return the last packagist version.
+     */
+    final protected function getLastPackageVersion(Package $package): Package\Version
```
OskarStark, line 113: Maybe "latest" is better than "last"?

src/Command/DispatchCommand.php
OskarStark: IMO we should keep the sonata-project.org website here

VincentLanglet: Why sonata-project over the specific bundle site ?
Should we update the one used on packagist then ?

VincentLanglet: @OskarStark Can you explain why ? :)

@sonata-project/contributors WDYT ?

src/Command/DispatchCommand.php
VincentLanglet: @OskarStark I don't know well the dispatch command.

How can I tried it locally without trigger any change on sonata projects ?

src/Command/DispatchCommand.php
OskarStark: I just feel its better to redirect the user to the mainpage of our project website

src/Command/AbstractCommand.php
OskarStark: Can you rename the method accordingly and remove the comment? @VincentLanglet

src/Command/AbstractCommand.php
```diff
@@ -106,4 +106,19 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
...
+    {
+        $versions = $package->getVersions();
+        $lastVersion = reset($versions);
+
+        if (false === $lastVersion) {
+            throw new \LogicException('The package has no version');
```
OskarStark, line 119: In this case we cannot update the settings of a new repo because this exception is not caught... 🤔

src/Command/DispatchCommand.php
```diff
@@ -141,12 +141,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
     private function updateRepositories(Package $package, array $projectConfig): void
     {
         $repositoryName = $this->getRepositoryName($package);
+        $lastVersion = $this->getLastPackageVersion($package);
```
OskarStark, line 144: Can we catch here and set `http://sonata-project.org`, as default?

And I would prefer to remove the trailing `/` from the url

src/Command/DispatchCommand.php
jordisala1991: Unless someone starts maintaining the website, redirecting to the mainpage feels like going to an unmaintained project (looks old and outdated). For me, going directly to the docs of the bundle is better

src/Command/AbstractCommand.php
VincentLanglet: Sure, i will try to find time for this !

src/Command/DispatchCommand.php
```diff
@@ -688,6 +694,8 @@ private function renderFile(Package $package, string $repositoryName, string $br
                     'package_title' => ucwords(str_replace(['-project', '/', '-'], ['', ' ', ' '], $package->getName())),
                     'package_description' => $package->getDescription(),
                     'packagist_name' => $package->getName(),
+                    'packagist_homepage' => $this->getLastPackageVersion($package)->getHomepage(),
+                    'packagist_keywords' => $this->getLastPackageVersion($package)->getKeywords(),
```
OskarStark, line 698: Do we really need them for the templates? 🤔

VincentLanglet: No indeed

src/Command/DispatchCommand.php
VincentLanglet: I do `'homepage' => $latestVersion->getHomepage() ?: 'https://sonata-project.org',`

isn't it enough ?

src/Command/AbstractCommand.php
VincentLanglet: Fixed, I use an empty version now.

PR #921, opened 2020-09-06
Multiple dev-kit improvement/fix

VincentLanglet:
I found multiple things I could improve
- The linter test was failing
- Some yml/yaml files were using 2 spaces instead as indent of 4.
- There was a weird thing in `.doctor-rst.yaml`

### DIFF
--- a/templates/project/.editorconfig
+++ b/templates/project/.editorconfig
@@ -14,10 +14,10 @@ trim_trailing_whitespace = true
 [*.{yaml,yml,twig,php}]
 indent_size = 4
 
-[*.{js,json,scss,css}]
-indent_size = 2
+[.yamllint]
+indent_size = 4
 
-[.{travis.yml,flintci.yml}]
+[*.{js,json,scss,css}]
 indent_size = 2
 
 [composer.json]
--- a/config/dev-kit.yaml
+++ b/config/dev-kit.yaml
@@ -1,58 +1,58 @@
 labels:
-  - name: patch
-    color: '207de5'
-  - name: minor
-    color: '009800'
-  - name: major
-    color: 'e11d21'
-  - name: pedantic
-    color: 'd1e4fa'
-  - name: enhancement
-    color: '02d7e1'
-  - name: feature
-    color: '02e10c'
-  - name: bug
-    color: 'fc2929'
-  - name: unconfirmed
-    color: '444444'
-  - name: critical
-    color: 'fc2929'
-  - name: vendor
-    color: 'ededed'
-  - name: pending author
-    color: 'ededed'
-  - name: stale
-    color: 'ededed'
-  - name: in progress
-    color: 'ededed'
-  - name: RTM
-    color: 'ffffff'
-  - name: docs
-    color: 'fbca04'
-  - name: help wanted
-    color: '0e8a16'
-  - name: Easy Pick
-    color: 'd7e102'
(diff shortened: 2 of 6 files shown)


templates/project/.editorconfig
```diff
@@ -14,10 +14,10 @@ trim_trailing_whitespace = true
 [*.{yaml,yml,twig,php}]
 indent_size = 4
 
-[*.{js,json,scss,css}]
-indent_size = 2
+[.yamllint]
```
core23, line 17:
```suggestion
[.yaml]
```

?

greg0ire: @core23 there really is a `.yamllint` file in YAML format, I think this is intended.

templates/project/docs/.doctor-rst.yaml.twig
```diff
@@ -40,9 +40,3 @@ whitelist:
 {% if custom_doctor_rst_whitelist_part is not empty %}
     {{ custom_doctor_rst_whitelist_part|raw }}
 {% endif %}
-{% elseif custom_doctor_rst_whitelist_part is not empty %}
-
-whitelist:
-    lines:
-    {{ custom_doctor_rst_whitelist_part|raw }}
-{% endif %}
```
greg0ire, line 48: What do you mean, "Fix conflict"?

VincentLanglet: There is something weird here:
```
if foo
endif
elseif foo
endif
```

Don't know if this is from a conflict, but I don't understand how it could work.

The twig syntax is wrong to me, and even if it wasn't, there will be to whitelist key.

templates/project/.editorconfig
VincentLanglet: yes it's intended

templates/project/docs/.doctor-rst.yaml.twig
greg0ire: It looks wrong indeed. So what you mean is that you are fixing a bad conflict resolution that happened prior to any of your commits.

> there will be to whitelist key.

Had to read that twice but okay, I get it!

greg0ire (dismissed):

templates/project/docs/.doctor-rst.yaml.twig
OskarStark: IMO the elseif needs to be if. Please let me double check it next week.

I am currently on a phone 📱 

I will merge it then

OskarStark: Or please create a separate PR for DOCtor file and merge the others right away. 👍🏻

OskarStark: @VincentLanglet the c
Code is correct.

The elseif belongs to the first if:

image

VincentLanglet: I revert it.

Damned, so tricky.

greg0ire (approved):

jordisala1991 (approved):

OskarStark (approved):

OskarStark:
Thanks 🙂

templates/project/docs/.doctor-rst.yaml.twig
OskarStark: Thanks 😊

PR #927, opened 2020-09-14
Enhancement: Use Action value object

OskarStark:


tests/Github/Domain/Value/Webhook/ActionTest.php
```diff
@@ -0,0 +1,143 @@
...
+        self::assertSame(
+            'synchronize',
+            Action::SYNCHRONIZE()->toString()
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function CEATED(): void
```
greg0ire, line 58: Should be `CREATED`

PR #940, opened 2020-09-14
Enhancement: Use Label value object

OskarStark:


src/Github/GithubClient.php
```diff
@@ -32,10 +33,10 @@ public function __construct(Client $client)
     /**
      * Adds a label from an issue if this one is not set.
      */
-    public function addIssueLabel(Repository $repository, IssueId $issueId, string $label): void
+    public function addIssueLabel(Repository $repository, IssueId $issueId, Label $label): void
     {
         foreach ($this->client->issues()->labels()->all($repository->username(), $repository->name(), $issueId->toInt()) as $labelInfo) {
-            if ($label === $labelInfo['name']) {
+            if ($label->equals(Label::fromString($labelInfo['name']))) {
```
greg0ire, line 39: That looks cumbersome, there should be a dedicated method in the VO in my opinion

OskarStark: Can explain? You meant o compare it with a string?

greg0ire: Yeah, and in general, anything you can do to make usage of a VO short and sweet is best. For the naming, `hasForValue()`, maybe?

OskarStark: Yes true, but in this case this is just the first step and having more typesafety instead using strings and check over and over again for not null/not empty ...

But yes, its just a first step 😄

PR #942, opened 2020-09-14
Enhancement: Add tests

OskarStark:
ping @sonata-project/core 

Did anyone of you tried to mock the Client?
```
1) App\Tests\Github\GithubClientTest::addIssueLabelAddsLabelIfItDoesNotExist
Trying to configure method "issues" which cannot be configured because it does not exist, has not been specified, is final, or is static
```

### DIFF
--- a/tests/Github/GithubClientTest.php
+++ b/tests/Github/GithubClientTest.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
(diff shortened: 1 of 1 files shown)


franmomu:
I think, you could do:
```php
$client = $this->getMockBuilder(Client::class)
    ->addMethods(['issues'])
    ->getMock();
```

wbloszyk:
I think PhpUnit don't see `@method`. With this Stub or works correct.

```php
namespace App\Tests\Stub;

use Github\Api\Issue;

class Client extends \Github\Client
{
    public function issues(): Issue
    {
        return parent::issues();
    }
}
```

PR #945, opened 2020-09-14
Enhancement: Extend label

OskarStark:


src/Github/Domain/Value/Label.php
```diff
@@ -21,20 +21,37 @@
...
         $this->value = $value;
+        $this->color = $color;
     }
 
     public static function fromString(string $value): self
     {
         return new self($value);
     }
 
+    public static function fromResponse(array $response): self
```
greg0ire, line 44: Please document any `array` parameter (so here it would be `array{name: string, color: string}`?)

wbloszyk: What about use jms_serializer and deserialize string directly from github API?

OskarStark: That could make sense

PR #947, opened 2020-09-15
Enhancement: Use constant instead of string

OskarStark:
Lets see if the YAML linter supports custom tags now....


OskarStark:
No 😢

PR #954, opened 2020-09-15
Enhancement: Inject Github client and paginator instead of initializing them in AbstractCommand

OskarStark:


src/Command/ReleaseCommand.php
```diff
@@ -45,12 +46,16 @@ final class ReleaseCommand extends AbstractCommand
     ];
 
     private Client $packagist;
+    private \Github\Client $githubClient;
```
jordisala1991, line 49: better add uses with alias

OskarStark: Will do it in a next PR

PR #965, opened 2020-09-15
Enhancement: Instantiate SymfonyStyle in initialize directly, instead in AbstractCommand

OskarStark:


OskarStark:
We need to keep it for now, as `AbstractNeedApplyCommand` need it too 🤔

PR #971, opened 2020-09-15
Bump phpstan to level 5

wbloszyk:


src/Command/AbstractCommand.php
```diff
@@ -23,7 +23,7 @@
 use Symfony\Component\Console\Style\SymfonyStyle;
 use function Symfony\Component\String\u;
 use Symfony\Component\Yaml\Yaml;
-
+use Symfony\Component\String\UnicodeString;
```
OskarStark, line 26: Is this needed?

wbloszyk: It's not needed. I run php-cs-fixer but some error from other files ddint allow fix it.

wbloszyk: You add properties type which is allowed from PHP 7.4, I try use 7.2 :smile:

OskarStark (approved):

PR #972, opened 2020-09-15
Enhancement: Upgrade sebastian/diff

OskarStark:
But I cannot find where it is used 🤔 

Bildschirmfoto 2020-09-15 um 11 56 26


greg0ire:
That requirement was added in https://github.com/sonata-project/dev-kit/pull/31, maybe it was locked at that version to avoid a bug? IMO you could try dropping it.

OskarStark:
> That requirement was added in #31, maybe it was locked at that version to avoid a bug? IMO you could try dropping it.

I will merge this one first and then try to remove it 👍

PR #1000, opened 2020-09-16
Spell "dev" properly

greg0ire:
Draft because I have to duplicate the env var.


greg0ire:
I redeployed, we now have two variables with the same value. After merging and deploying this, we should remove the old variable.

VincentLanglet (approved):

OskarStark:
WTF, never noticed it 😄  Thanks!

greg0ire:
I deployed, and removed the env var, and now I'm redeploying for that to take effect

PR #1019, opened 2020-09-17
Enhancement: Set topics via Github API

OskarStark:


src/Command/Dispatcher/DispatchSettingsCommand.php
```diff
@@ -58,7 +60,12 @@
             try {
                 $this->io->title($project->name());
 
+                if ('doctrine-phpcr-admin-bundle' !== $project->name()) {
```
VincentLanglet, line 63: Why this check ?

OskarStark: For testing 😂

OskarStark: Will remove it later when the sorting is fixed. Have a short appointment now

PR #1092, opened 2020-09-23
Enhancement: Use faker for tests

OskarStark:


wbloszyk:
What about using [AliceBundle](https://github.com/hautelook/AliceBundle) ?  I know alice use this faker.

OskarStark:
> What about using AliceBundle ? I know alice use this faker.

We have no entities which needs to be populated 🤔

localheinz (commented):
👍

PR #1140, opened 2020-10-02
Enhancement: Add badge for type coverage

OskarStark:
Looks like it is always the latest, so we cannot have 2 different badges for stable/unstable branch 🤔


core23:
It is using the default branch (our stable), which should be okay
https://github.com/psalm/shepherd/blob/7d9fd2a82eedad3f40f0ffed5922bc646eb8b53c/src/GithubApi.php#L18

core23 (approved):

VincentLanglet (approved):

PR #1148, opened 2020-10-04
Use sonata-project/admin-bundle 4.x in sonata-project/translation-bundle

franmomu:
Ref: https://github.com/sonata-project/SonataTranslationBundle/pull/383

### DIFF
--- a/config/projects.yaml
+++ b/config/projects.yaml
@@ -415,10 +415,10 @@ timeline-bundle:
 translation-bundle:
     branches:
         master:
-            php: ['7.2', '7.3', '7.4']
+            php: ['7.3', '7.4']
             variants:
                 symfony/symfony: ['4.4']
-                sonata-project/admin-bundle: ['3']
+                sonata-project/admin-bundle: ['dev-master']
         2.x:
             php: ['7.2', '7.3', '7.4']
             variants:


config/projects.yaml
```diff
@@ -415,10 +415,10 @@ timeline-bundle:
 translation-bundle:
     branches:
         master:
-            php: ['7.2', '7.3', '7.4']
+            php: ['7.3', '7.4']
             variants:
                 symfony/symfony: ['4.4']
-                sonata-project/admin-bundle: ['3']
+                sonata-project/admin-bundle: ['4.x-dev']
```
jordisala1991, line 421: better use dev-master here, to do the same than on doctrine orm admin

franmomu: why is it better? just out of curiosity

jordisala1991: just consistency and ensure the twig filter that @VincentLanglet implemented works the same way

VincentLanglet: Cf: https://github.com/sonata-project/dev-kit/blob/master/src/Twig/Extension/GithubExtension.php#L19
And https://github.com/sonata-project/dev-kit/blob/master/templates/project/.github/workflows/test.yaml.twig#L53

But maybe there is something better to do ? 

Without any change, we will try to run test with the version `'4.x-dev.x'`

OskarStark (approved):

PR #1177, opened 2020-10-14
Add psalm to sonata-project/doctrine-mongodb-admin-bundle

franmomu:
Ref: https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/446

EDIT: looks like the mongodb extension should be installed, I'll take a look later today.

### DIFF
--- a/config/projects.yaml
+++ b/config/projects.yaml
@@ -190,6 +190,7 @@ doctrine-extensions:
 
 doctrine-mongodb-admin-bundle:
     phpstan: true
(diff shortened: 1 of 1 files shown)


greg0ire (approved):

franmomu:
We cannot run psalm in the packages which need php extensions not installed by default... should we just use `psalm` directly since it's in the dev requirements instead of the https://github.com/psalm/psalm-github-actions?

PR #1196, opened 2020-10-21
Remove specific PHPUnit flow for php 8

jordisala1991:
This was previously done because prophecy was not compatible with php8. Now that it is compatible, the workflow wont stop at the install point of simple-phpunit execution. Simplify and unify workflow is good: POC: https://github.com/sonata-project/SonataMediaBundle/pull/1844

### DIFF
--- a/templates/project/.github/workflows/test.yaml.twig
+++ b/templates/project/.github/workflows/test.yaml.twig
@@ -86,7 +86,12 @@ jobs:
 {% endverbatim %}
       - name: Configuration required for PHP 8.0
         if: matrix.php-version == '8.0'
-        run: composer config platform.php 7.4.99 && composer require phpunit/phpunit:"9.3.*" --no-update
+        run: composer config platform.php 7.4.99
+
+      - name: Sets phpunit 9 for PHP 8.0
+        if: matrix.php-version == '8.0'
+        run: |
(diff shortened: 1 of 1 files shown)


templates/project/.github/workflows/test.yaml.twig
```diff
@@ -86,7 +86,12 @@ jobs:
 {% endverbatim %}
       - name: Configuration required for PHP 8.0
         if: matrix.php-version == '8.0'
-        run: composer config platform.php 7.4.99 && composer require phpunit/phpunit:"9.3.*" --no-update
+        run: composer config platform.php 7.4.99
+
+      - name: Sets phpunit 9 for PHP 8.0
```
jordisala1991, line 91: Once all the repositories are fully compatible with phpunit 9 (there are still some issues like: https://github.com/sonata-project/sonata-doctrine-extensions/pull/255, and the removal of prophecy from our code base) We will change this to a condition for php 7.2 (if it is present on php versions) to use PHPUnit 8, and the rest of the builds will use PHPunit 9

Once php 7.2 is completely removed from maintained branches, we can fully remove this. (so less code to maintain)

jordisala1991:
The failure on the linter does not seem related to this PR @OskarStark

OskarStark (approved):
Looks good to me 👍

PR #1203, opened 2020-10-21
Fix cyclic dependency on QA

jordisala1991:


jordisala1991:
Proof PR: https://github.com/sonata-project/SonataBlockBundle/pull/767

... It does not work for phpstan, but it does for psalm..

OskarStark (approved):

OskarStark:
> ... It does not work for phpstan, but it does for psalm..

Hmm, if it does not work for PHPStan, shall we just add it for Psalm now?

jordisala1991:
Yes, but it won't fully solve the issue.

jordisala1991:
Fixed only the psalm now.

PR #1206, opened 2020-10-25
Fix block-bundle

jordisala1991:
Proof: https://github.com/sonata-project/SonataBlockBundle/pull/773

### DIFF
--- a/templates/project/.github/workflows/qa.yaml.twig
+++ b/templates/project/.github/workflows/qa.yaml.twig
@@ -23,6 +23,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+{# Remove when removing 3.x for the managed branches of block-bundle #}
+{% if project.repository.name == 'SonataBlockBundle' and branch.name == '3.x' %}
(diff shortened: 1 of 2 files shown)


jordisala1991:
This reveals the true problem with php 8: tracked here: https://github.com/sonata-project/dev-kit/issues/1207

templates/project/.github/workflows/qa.yaml.twig
```diff
@@ -23,6 +23,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+{# Remove when removing 3.x for the managed branches of block-bundle #}
+{% if project.repository.name == 'SonataBlockBundle' and branch.name == '3.x' %}
```
VincentLanglet, line 27: Maybe it's better to remove the check, to add this environment variable in every build
since I need it too: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1170

jordisala1991: IMO on the doctrineOrm we should first see why it is needed

OskarStark (approved):

PR #1213, opened 2020-10-28
Stop using custom action for static analysis

VincentLanglet:
See the POC: https://github.com/sonata-project/SonataAdminBundle/pull/6549

Should we run static analysis on multiple PHP version, and lowest/highest dependency like I did ?
It's interesting, some of them are failing.

Or maybe just two build:
- Lowest PHP version, lowest dependencies.
- Highest PHP version, highest dependencies.

### DIFF
--- a/templates/project/.github/workflows/documentation.yaml.twig
+++ b/templates/project/.github/workflows/documentation.yaml.twig
@@ -5,60 +5,60 @@
 name: Documentation
 
 on:
-  push:
-    branches:
+    push:
+        branches:
 {% for branch in project.branchesReverse %}
-      - {{ branch.name }}
+            - {{ branch.name }}
 {% endfor %}
-  pull_request:
+    pull_request:
 
 jobs:
-  build:
-    name: Sphinx build
+    build:
+        name: Sphinx build
 
-    runs-on: ubuntu-latest
+        runs-on: ubuntu-latest
 
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2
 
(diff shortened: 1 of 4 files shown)


jordisala1991:
I would say only one run of static analysis

VincentLanglet:
> I would say only one run of static analysis

I would have say this too.

But then I thought about these cases:

If we require the library `foo: ^3.4`, which has a new version 3.5.
If this library provide the function
```
bar(string $string) () {}
```
Which was changed to 
```
bar(?string $string) () {}
```
in 3.5.

Calling `bar(null)` won't report an error in our current build ; but would in 3.4.

It would help us to know what we don't support `foo: ^3.4` but `foo: ^3.5` instead.

jordisala1991:
> > I would say only one run of static analysis
> 
> I would have say this too.
> 
> But then I thought about these cases:
> 
> If we require the library `foo: ^3.4`, which has a new version 3.5.
> If this library provide the function
> 
> ```
> bar(string $string) () {}
> ```
> 
> Which was changed to
> 
> ```
> bar(?string $string) () {}
> ```
> 
> in 3.5.
> 
> Calling `bar(null)` won't report an error in our current build ; but would in 3.4.
> 
> It would help us to know what we don't support `foo: ^3.4` but `foo: ^3.5` instead.

Yes, but it is already a lot of work keeping green those static analisis builds, imagine if we need to test with multiple php versions / dependencies. IMO it is not worth it. We should ensure that the code is green with phpstan on latests versions of everything. The lowest builds and compat are already tested with phpunit, so the compatibility is guaranteed

VincentLanglet:
Done then

OskarStark (approved):

PR #1226, opened 2020-10-29
Enhancement: Use specific ubuntu version

OskarStark:
at the end of november this woould otherwise point to 20.04

Based on https://github.com/twigphp/Twig/pull/3419#discussion_r514367410

For more info: https://github.com/shivammathur/setup-php#cloud-osplatform-support


OskarStark:
Based on this comment: https://github.com/twigphp/Twig/pull/3419#discussion_r514404417 it is not really needed, but not wrong at all. Lets be specific, as it is easy for us to update a version via dev-kit

PR #1231, opened 2020-11-01
Add verbatim tag when using dollar symbol

franmomu:
They were missing in https://github.com/sonata-project/dev-kit/pull/1213

Ref: https://github.com/sonata-project/SonataAdminBundle/blob/3f0f8dd287e0d3e8a4a599d4f9cddb7a10688bbb/.github/workflows/qa.yaml#L71


OskarStark (approved):
Nice catch

PR #1247, opened 2020-11-05
Fix: Use html_url instead of api url

OskarStark:
CleanShot 2020-11-05 at 14 47 19


OskarStark:
Works 💪 

CleanShot 2020-11-05 at 14 50 48

PR #1298, opened 2020-12-01
Use PHP 7.4 for target php in sonata-project/intl-bundle

franmomu:
sonata-project/user-bundle is not compatible with PHP 8.

@OskarStark not sure about this but I think that setting `target_php` would be the PHP version to use in variants, is that right?


OskarStark:
Yes it is

OskarStark:
But we should of course make user bundle work with php 8 and then switch back here 👌🏻

PR #1330, opened 2021-01-02
Clean up phpunit configuration

franmomu:
Some time ago I started this: https://github.com/franmomu/SonataUserBundle/pull/1 (POC)

This PR does 3 things:

- Removed PHPUnit attributes that are not necessary because those values are the default ones.
- Added `failOnRisky`.
- Removed some legacy code and add an extension point for custom boostraping.

I think `autoload.php.dist` was used in older Symfony versions and the `try to get Symfony's PHPUnit Bridge` part I think it is not necessary anymore.


OskarStark:
If you want to delete the file you first need to rename it to: `*_DELETE` and in a second PR we need to remove it completely, otherwise dev-kit would just delete the file and would not compare for the diff.

OskarStark (approved):

OskarStark:
I deploy, please keep an open 👁️  on the PR's, thank you!

franmomu:
> If you want to delete the file you first need to rename it to: `*_DELETE` and in a second PR we need to remove it completely, otherwise dev-kit would just delete the file and would not compare for the diff.
> 
> I will do this for you and merge and deploy 👍
> 
> Thanks

Oh true 🤦 , thank for doing it!

OskarStark:
Ah forget the pings, you removed the processIsolation config 👌🏻👍🏻

PR #1331, opened 2021-01-03
Fix typo and add panther

franmomu:
I added a typo in `custom_bootstrap.php`.

This also adds a `panther` option related to https://github.com/symfony/panther, with this library we can make functional tests easier.

I also added ORM bundle for https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1242


config/projects.yaml
```diff
@@ -193,11 +193,7 @@ doctrine-extensions:
 doctrine-mongodb-admin-bundle:
     phpstan: true
     psalm: true
-    custom_gitignore_part: |
-        /tests/App/public/bundles
-    excluded_files:
-        - phpunit.xml.dist.twig
-        - tests/bootstrap.php.twig
```
VincentLanglet, line 200: If I understand correctly, a PR is needed for doctrine-mongodb-admin-bundle, to add a custom_boostrap.

templates/project/.gitignore.twig
```diff
@@ -13,6 +13,9 @@ phpstan.neon
 {% if project.hasDocumentation %}
 /docs/_build/
 {% endif %}
+{% if project.usesPanther %}
+/tests/App/public/bundles
```
VincentLanglet, line 17: This is related to https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/blob/3.x/tests/bootstrap.php#L52-L53
```
$publicDir = __DIR__.'/App/public';
$_SERVER['PANTHER_WEB_SERVER_DIR'] = $publicDir;
```

So I would recommend to add these two lines in the boostrap.php file (and not the custom one) if the panther option is set to true.

OskarStark (approved):

config/projects.yaml
franmomu: Yes, but as it is now `custom_boostrap` instead of `custom_bootstrap` I was going to wait this to be deployed and I'll do it in that PR.

templates/project/.gitignore.twig
franmomu: I added the variable as a server variable in phpunit.xml.dist

OskarStark (approved):

OskarStark:
So ready to merge and deploy?

franmomu:
> So ready to merge and deploy?

I think it is.

VincentLanglet (approved):

PR #1098, opened 2020-09-23
Test - automatic CS fix + commit (in a PR from forked repo)

OskarStark:
Lets see if a PR receives an automatic CS fix

PR #1332, opened 2021-01-03
POC

franmomu:
To test https://github.com/sonata-project/dev-kit/pull/1098, just in case it works now


franmomu:
Nope 😞 (I just tried because it worked, but in my branch, not with a remote one)

PR #1341, opened 2021-01-04
Add cache back

VincentLanglet:
Blockbundle still require the `cache` library. So we should keep the CI (especially for the php8 build)

### DIFF
--- a/config/projects.yaml
+++ b/config/projects.yaml
@@ -124,9 +124,9 @@ cache:
   docs_target: false
   branches:
     master:
-      php: ['7.3', '7.4']
+      php: ['7.3', '7.4', '8.0']
     2.x:
-      php: ['7.3', '7.4']
+      php: ['7.3', '7.4', '8.0']
 
 classification-bundle:
     custom_doctor_rst_whitelist_part:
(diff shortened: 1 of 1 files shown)


config/projects.yaml
```diff
@@ -118,6 +118,16 @@ block-bundle:
             variants:
                 symfony/symfony: ['4.4']
 
+cache:
+  excluded_files:
+    - README.md
+  docs_target: false
+  branches:
+    master:
+      php: ['7.3', '7.4']
```
core23, line 127: No php8?

OskarStark, line 127:
```suggestion
      php: ['7.3', '7.4', '8.0']
```

config/projects.yaml
```diff
@@ -118,6 +118,16 @@ block-bundle:
...
+    2.x:
+      php: ['7.3', '7.4']
```
OskarStark, line 129:
```suggestion
      php: ['7.3', '7.4', '8.0']
```

config/projects.yaml
OskarStark: Added, thanks!

VincentLanglet:
The php 8 build shouldnt be added as mandatory. https://github.com/sonata-project/dev-kit/pull/1342

PR #1346, opened 2021-01-05
Add apcu for cache repository

VincentLanglet:
I think the `ext-apcu` extension is needed for the cache library and the cache-bundle one.

I'm trying to add it.

### DIFF
--- a/templates/project/.github/workflows/test.yaml.twig
+++ b/templates/project/.github/workflows/test.yaml.twig
@@ -71,6 +71,8 @@ jobs:
 {% if branch.services|length > 0 %}
                   tools: composer:v{{ project.composerVersion }}, pecl
                   extensions: {{ branch.services|split(', ') }}
+              env:
+                  fail-fast: true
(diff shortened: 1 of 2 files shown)


templates/project/.github/workflows/test.yaml.twig
```diff
@@ -68,10 +68,11 @@ jobs:
               with:
                   {% verbatim %}php-version: ${{ matrix.php-version }}{% endverbatim %}
                   coverage: pcov
-{% if branch.hasService('mongodb') %}
```
OskarStark, line 71: hmm it looks ok, but I think this hasService('mongo') is used somewhere 🤔

VincentLanglet: What do you mean ?

OskarStark: did you search the whole project for "hasService('mongodb')"? Makefiles etc?

PR #1360, opened 2021-01-19
Install webdrivers when using panther

franmomu:
See https://github.com/symfony/panther/blob/19c317132b078d243b69b08aafc466d7b6326826/README.md#installing-chromedriver-and-geckodriver

POC: https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/512

We have to bump `symfony/panther` to `0.9`, but that can be done in the `dev-kit` PR.

PR #1404, opened 2021-02-04
Add drivers folder to gitignore if panther is used

VincentLanglet:
Since the version `0.9` of panther 
```
composer require --dev dbrekelmans/bdi
vendor/bin/bdi detect drivers
```
is required

We need to ignore the folder `drivers` then.

Any idea how to run `vendor/bin/bdi detect drivers` automatically too ?

### DIFF
--- a/templates/project/.gitignore.twig
+++ b/templates/project/.gitignore.twig
@@ -14,6 +14,7 @@ phpstan.neon
 /docs/_build/
 {% endif %}
 {% if project.usesPanther %}
+/drivers
 /tests/App/public/bundles
 {% endif %}
 {% if project.customGitignorePart is not empty %}


franmomu:
I created https://github.com/sonata-project/dev-kit/pull/1360, but as it looked like it worked without it, I closed it.

VincentLanglet:
@franmomu I get the error locally, so I think it's still needed.

And it shouldn't be added to the CI but directly to the composer.json or the developer can't test locally.

franmomu:
yeah, but you can add the `name: Download WebDriver drivers` part to CI.

VincentLanglet:
> yeah, but you can add the `name: Download WebDriver drivers` part to CI.

If we don't find a better solution, it could be inside the `make test` command or we could add a `make pretest`.
We have multiple solutions for the CI.

I was more concern for the first time contributor who try to run test and get an error with `vendor/bin/simple-phpunit tests`.
That's why I was wondering if we could run something in the custom_bootstrap.php or something similar.

OskarStark:
Any news here? Shall we merge this?

VincentLanglet:
> Any news here? Shall we merge this?

IMHO we should

OskarStark (approved):

PR #1413, opened 2021-02-13
Add entityAuditBundle

VincentLanglet:


config/projects.yaml
```diff
@@ -284,6 +284,16 @@ ecommerce:
             variants:
                 symfony/symfony: ['4.4']
 
+entity-audit-bundle:
+    excluded_files:
+        - LICENSE
```
OskarStark, line 289: Shouldn't we update the license?

VincentLanglet:
@OskarStark Are you available to try this ? :)

config/projects.yaml
VincentLanglet: I'm not sure, I prefer to discuss about it before. We can remove this line later if needed

PR #1415, opened 2021-02-14
Refactor github actions

VincentLanglet:
- Use 4 spaces for indent
- Fix the failing phpcs tests

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,82 +1,71 @@
 name: Lint
 
 on:
-  push:
-    branches:
-      - master
-  pull_request:
-  pull_request_target:
+    push:
+        branches:
+            - master
+    pull_request:
 
 jobs:
-  php-cs-fixer:
-    name: PHP-CS-Fixer
+    php-cs-fixer:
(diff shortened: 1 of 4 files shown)


VincentLanglet:
I'm not sure why the 7.4 test is not running and why I still have the phpcs pull_request_target test.
Does it require to be merged in order to see if it works ?

.github/workflows/auto-merge.yaml
```diff
@@ -0,0 +1,42 @@
...
+    pull_request:
+
+jobs:
+    auto-merge:
+        name: Auto-Merge
+
+        runs-on: ubuntu-latest
+
+        needs:
+            - test
```
OskarStark, line 16: This does not work, it only knows about jobs in the same workflow

OskarStark:
If it comes from a fork yes

.github/workflows/auto-merge.yaml
VincentLanglet: Didn't know this, I rollback the split.

OskarStark (approved):

OskarStark (approved):

PR #1427, opened 2021-02-23
Remove yoda conditions

franmomu:
Remove yoda conditions, we have (:trollface:)

What about this?

PS: IMHO we could change to Doctrine CS, but that's another topic.

### DIFF
--- a/src/Action/DetermineNextRelease.php
+++ b/src/Action/DetermineNextRelease.php
@@ -76,7 +76,7 @@ final class DetermineNextRelease
             Query::pullRequestsSince($repository, $branch, $currentRelease->publishedAt(), AbstractCommand::SONATA_CI_BOT)
         );
 
-        if ([] === $pullRequests) {
+        if ($pullRequests === []) {
             throw NoPullRequestsMergedSinceLastRelease::forProject(
                 $project,
                 $currentRelease->publishedAt()
--- a/src/Action/DetermineNextReleaseVersion.php
+++ b/src/Action/DetermineNextReleaseVersion.php
@@ -24,7 +24,7 @@ final class DetermineNextReleaseVersion
      */
     public static function forTagAndPullRequests(Tag $current, array $pullRequests): Tag
     {
-        if ([] === $pullRequests) {
+        if ($pullRequests === []) {
             return $current;
         }
 
--- a/src/Command/AutoMergeCommand.php
+++ b/src/Command/AutoMergeCommand.php
@@ -60,7 +60,7 @@ final class AutoMergeCommand extends AbstractNeedApplyCommand
         $projects = $this->projects->all();
 
         $title = 'Merge branches of repositories if there is no conflict';
-        if ([] !== $input->getArgument('projects')) {
+        if ($input->getArgument('projects') !== []) {
             $projects = $this->projects->byNames($input->getArgument('projects'));
             $title = sprintf(
                 '%s for: %s',
@@ -100,7 +100,7 @@ final class AutoMergeCommand extends AbstractNeedApplyCommand
         // Merge the oldest branch into the next newest, and so on.
         while (($head = current($branchNames))) {
             $base = next($branchNames);
-            if (false === $base) {
+            if ($base === false) {
                 break;
             }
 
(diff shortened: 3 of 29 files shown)


OskarStark:
I am curious, why?

greg0ire (approved):
@OskarStark https://dev.to/greg0ire/why-using-yoda-conditions-you-should-probably-not

greg0ire:
> IMHO we could change to Doctrine CS, but that's another topic.

That would be a great move IMO

VincentLanglet (changes requested):
The yoda condition is in the symfony standard https://symfony.com/doc/current/contributing/code/standards.html#structure

I think it's better to stick to this instead of creating a new standard.

franmomu:
I was moved to create this PR (apart from the comments @greg0ire said) with this suggestion: https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/537#discussion_r580517462

`if (null === $data['value'] ?? null) {`

that has to be like this (and is not easy to read):

`if (null === ($data['value'] ?? null)) {`

I don't think is creating a new standard, is modifying a controversial part [that Symfony may not change it because of the amount of work](https://github.com/symfony/symfony-docs/issues/12965). But maybe it's better to propose to change to Doctrine CS instead of modify yoda conditions.

greg0ire:
IMO the sf standard has one big flaw: there are a few rules that they are not 100% sure about that stil they keep around because it would be to painful to do the change and then merge up. IMO there should be a Symfony standard (rules they are really proud of and want to recommend to the community), and a Symfony internal standard (rules that they need to enforce in the codebase).

core23:
IMHO we should stick with the symfony cs.

> `if (null === $data['value'] ?? null) {`

This kind of code should be avoided in any case not matter what coding style we are using.

This is also dangerous when PHP is evolving. They did change the execution order recently for a similar syntax.

OskarStark:
If this can be done automatically fine, otherwise we have other things to do and not enough man power, same like for `master` to `main` and remove `offensive wording`, I agree we can and maybe should, but I think we can't 😢

phansys:
I agree about following the standard proposed by Symfony.
About the expression, `null === $data['value'] ?? null`, I think it can be safely replaced with `!isset($data['value'])`.

wbloszyk:
> About the expression, `null === $data['value'] ?? null`, I think it can be safely replaced with `!isset($data['value'])`.

Do not agree. What if it `$data['value'] = null`?

VincentLanglet:
All the coding CS are opinionated.
- The Symfony one has Yoda comparison and no space between concatenation for instance (which is not PSR12).
- I think Doctrine has a space between method return type and a space after the `!` operator.

I think if we start to follow a CS, we should not override rules from it. We can add rules, but not override somes.
It's not the best for a contributor to work with a project A with Symfony CS, a project B with Symfony CS without Yoda, a project C with Symfony CS with Yoda but with space for concatenation, and so on.

So I'm fine to stick with Symfony CS.

> But maybe it's better to propose to change to Doctrine CS instead of modify yoda conditions.

A POC PR on SonataAdminBundle would show all the differences between the CS.
But does the Doctrine CS works with phpcs fixer ?

> > About the expression, `null === $data['value'] ?? null`, I think it can be safely replaced with `!isset($data['value'])`.
> 
> Do not agree. What if it `$data['value'] = null`?
> Then `!isset($data['value']) || $data['value'] !== null` should be used.

If `$data['value'] = null`, isset return false.

phansys:
>But does the Doctrine CS works with phpcs fixer ?

AFAIK, it does not provide integration with PHP-CS-Fixer: https://github.com/doctrine/coding-standard.

greg0ire:
Doctrine CS integrates with phpcs, which comes with 2 binaries: phpcs and phpcbf. That's because all issues cannot be fixed automatically (which means Doctrine CS is stricter).

Regarding the overrides, the rules that I don't like this much with Doctrine is you have to align equal signs. With Symfony, I don't like Yoda, multiline signatures are forbidden, and multiline exceptions are forbidden too.

For multiline signatures, I tried to propose a change and it was rejected because hard to merge up: https://github.com/symfony/symfony/issues/18890#issuecomment-221874231

VincentLanglet:
> Doctrine CS integrates with phpcs, which comes with 2 binaries: phpcs and phpcbf. That's because all issues cannot be fixed automatically (which means Doctrine CS is stricter).

This also means that the issues are harder to be fixed for someone who doesn't use often linters.
I had multiples times developers in my teams asking how to fix a phpcs issue. 

Anyway, the best way to discuss about the doctrine CS standard would be a PR on sonataAdmin first.

franmomu:
> I think if we start to follow a CS, we should not override rules from it. We can add rules, but not override somes.

By the way, we are already overriding `single_line_throw` :trollface:

PR #1453, opened 2021-03-10
Fix yaml

VincentLanglet:


phansys (commented):
I thought it was a new syntax that I didn't know about.

VincentLanglet:
We should add test for this. I'll try when I'll have time.

phansys:
Maybe `bin/console lint:yaml` is enough.

PR #1459, opened 2021-03-10
Add zip extensions

VincentLanglet:
According to the dev-kit ci, it's needed.

### DIFF
--- a/config/projects.yaml
+++ b/config/projects.yaml
@@ -279,10 +279,12 @@ ecommerce:
     branches:
         master:
             php: ['7.3', '7.4']
+            php_extensions: ['zip']
             variants:
                 symfony/symfony: ['4.4']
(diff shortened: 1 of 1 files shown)


phansys (approved):
Sounds logic, but could you please tell how the dev-kit indicates about that requirement?

wbloszyk:
> Sounds logic, but could you please tell how the dev-kit indicates about that requirement?

Dev-kit will tell github action how to configure php instalation on docker. For now some PHP extensions are missing only  in PHP 7.3

VincentLanglet:
> Sounds logic, but could you please tell how the dev-kit indicates about that requirement?

https://github.com/sonata-project/SonataFormatterBundle/pull/561/checks?check_run_id=2080403527

PR #1472, opened 2021-03-23
Add .symfony.bundle file

VincentLanglet:
I'm trying to reproduce: https://github.com/EasyCorp/EasyAdminBundle/blob/master/.symfony.bundle.yaml


OskarStark:
I would propose to add the comments from the file, WDYT?

greg0ire (approved):

VincentLanglet:
> I would propose to add the comments from the file, WDYT?

Added.

I don't think we can use the `dev_branch_alias` key.
And I'm not sure about the `dev_branch` key. What do you understand, should it be master or the current branch ?

OskarStark:
@javiereguilaz can you please help us out here?

javiereguiluz:
For reference, the current config we use for Sonata Docs at symfony.com is:

```yaml
branches: ["2.x", "3.x", "master"]
maintained_branches: ["3.x", "master"]
doc_dir: { 2.x: "Resources/doc/", 3.x: "docs/", master: "docs/" }
current_branch: "3.x"
dev_branch: "master"
dev_branch_alias: "4.x"
```

Be careful about the `doc_dir` option ... you changed this in every version, so it's very important to define the right value for each version or docs won't be built. Thanks!

The "dev_branch" is the branch you use to work on the versions that will be released in the future. E.g., for Symfony, `5.2` is the current branch and `5.x` is the dev branch (and 5.3 is the dev_branch_alias because that's the version that future releases will use, instead of 5.x).

The "dev_branch_alias" is the version which will be used in the future when stable versions will be released from the currently "dev branch" (which will be the "current branch" in the future). Given that Sonata released 2.x and 3.x, the next version is probably 4.x (hence the `dev_branch_alias: "4.x"`) but maybe you want to name it `3.1`, `3.2`, et.c (like Symfony does). Any value is fine, including not defining this value, but it's nice to define it to avoid calling this version "master" or "main".

I hope it's more clear now.

VincentLanglet:
> For reference, the current config we use for Sonata Docs at symfony.com is:
> 
> ```yaml
> branches: ["2.x", "3.x", "master"]
> maintained_branches: ["3.x", "master"]
> doc_dir: { 2.x: "Resources/doc/", 3.x: "docs/", master: "docs/" }
> current_branch: "3.x"
> dev_branch: "master"
> dev_branch_alias: "4.x"
> ```

So if we're dropping the doc for 2.x, it becomes
```
branches: ["3.x", "master"]
maintained_branches: ["3.x", "master"]
doc_dir: "docs/"
current_branch: "3.x"
dev_branch: "master"
dev_branch_alias: "4.x"
```

We need a way to get the dev branch alias @OskarStark

OskarStark:
> We need a way to get the dev branch alias @OskarStark

Lets make in configurable, WDYT?

VincentLanglet:
> > We need a way to get the dev branch alias @OskarStark
> 
> Lets make in configurable, WDYT?

I think we can make a function to compute the alias.

OskarStark:
Then lets do it like this in the `Project` value object

templates/project/.symfony.bundle.yaml.twig
```diff
@@ -0,0 +1,47 @@
+# DO NOT EDIT THIS FILE!
+#
+# It's auto-generated by sonata-project/dev-kit package.
+
+# MANDATORY
+# An array with the name of the branches for which you want to publish docs
+# It doesn't have to be an exhaustive list of your repository branches, but
+# only the main repository branches (past or present)
```
greg0ire, line 8: Should we make the comments Twig comments instead of YAML comments? That way they will not be copied in every repository, I think they are only useful when you edit the file, and having them here contradicts the first comment about how this file should not be edited directly on repositories.

OskarStark: I like the idea

OskarStark:
open to finish this PR @VincentLanglet ?

We should merge it and see if everything is working fine on the `docs.symfony.com` side (cc @javiereguiluz)

VincentLanglet:
> open to finish this PR @VincentLanglet ?
> 
> We should merge it and see if everything is working fine on the `docs.symfony.com` side (cc @javiereguiluz)

I changed the comment.

I think we can merge this like this for now.

templates/project/.symfony.bundle.yaml.twig
```diff
@@ -0,0 +1,59 @@
+# DO NOT EDIT THIS FILE!
+#
+# It's auto-generated by sonata-project/dev-kit package.
```
OskarStark, line 3: This needs to be a twig comment too

PR #1473, opened 2021-03-23
Disable phpdoc_to_comment

VincentLanglet:
See https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4446

### DIFF
--- a/templates/project/.php_cs.dist
+++ b/templates/project/.php_cs.dist
@@ -47,6 +47,7 @@ $rules = [
     'nullable_type_declaration_for_default_null_value' => ['use_nullable_type_declaration' => true],
     'ordered_class_elements' => true,
     'ordered_imports' => true,
+    'phpdoc_to_comment' => false, // https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4446
     'php_unit_method_casing' => false,
     'php_unit_set_up_tear_down_visibility' => true,
     'php_unit_strict' => true,


franmomu (changes requested):
The `phpdoc_to_comment` is in the Symfony Standard https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.18/doc/ruleSets/Symfony.rst, shouldn't we stick to it and do not create a new standard?

VincentLanglet:
> The `phpdoc_to_comment` is in the Symfony Standard https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.18/doc/ruleSets/Symfony.rst, shouldn't we stick to it and do not create a new standard?

I will be happy to turn on the `phpdoc_to_comment` on back as soon as it will allow the `@psalm-suppress` tag.
Psalm tag are requiring a phpdoc.
You can look at
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4446.
- https://github.com/vimeo/psalm/issues/2344
There is a PR: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4649, but who knows when it will be merged.

It's a blocking point for https://github.com/sonata-project/SonataAdminBundle/pull/6961

jordisala1991:
But, what if instead of using the phpdoc you silence the errors using the phpstan baseline?

VincentLanglet:
> But, what if instead of using the phpdoc you silence the errors using the phpstan baseline?

When I look at the doc https://psalm.dev/docs/running_psalm/dealing_with_code_issues/, the baseline is presented like a list of errors that we currently don't fix but we want to fix later ; like a todo.

For instance here: https://github.com/sonata-project/SonataAdminBundle/pull/6961/files#diff-f451512366833b7ca7e185eb81f569d83e5d19defb248555e8e9a8c09495e241, the error is made on purpose and will never be fixed.

As a reference, psalm code is using these annotation: https://github.com/vimeo/psalm/blob/10f2966dcbb35e4d41ae3bd5482cf1cb17671bda/src/Psalm/Internal/LanguageServer/ProtocolStreamReader.php#L52

franmomu:
I've been facing this issue for a while and I end up adding them in the config file as in https://github.com/vimeo/psalm/blob/10f2966dcbb35e4d41ae3bd5482cf1cb17671bda/psalm.xml.dist#L45

VincentLanglet:
> I've been facing this issue for a while and I end up adding them in the config file as in https://github.com/vimeo/psalm/blob/10f2966dcbb35e4d41ae3bd5482cf1cb17671bda/psalm.xml.dist#L45

I will try this

PR #1496, opened 2021-04-15
Fixed the doc_dir option value

javiereguiluz:
SonataAdminBundle uses `docs/`, so we need to update this value. I don't know if it's the same for all Sonata bundles 🙏


jordisala1991:
Yes it is the same for all Sonata bundles

jordisala1991 (approved):

franmomu (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #1497, opened 2021-04-16
Run PHP 8 tests for user bundle

4c0n:
Run PHP 8 tests for user bundle.


VincentLanglet (changes requested):
Currently block by https://github.com/sonata-project/SonataUserBundle/pull/1341

PR #1505, opened 2021-04-30
Add default configuration when using assets

jordisala1991:
Some changes:

* Add new option to customize gitignore per branch
* Add new option: `assets` on a branch
* Add basic files to standarize the linting on all repos that have assets
* Add github workflow

Part of https://github.com/sonata-project/SonataAdminBundle/issues/7049


jordisala1991:
Example output for admin-bundle

```diff
 [WARNING] This is a dry run execution. No change will be applied here.


Dispatch files for: admin-bundle
================================

admin-bundle
------------

Files for 3.x
-------------

Delete /.babelrc.json file!
Delete /.eslintrc.json file!
Delete /.stylelintrc.json file!
Delete /.github/workflows/frontend.yaml file!
diff --git a/.gitignore b/.gitignore
index c3c28de3f..c66416c45 100644
--- a/.gitignore
+++ b/.gitignore
@@ -69,4 +69,3 @@ src/Resources/public/vendor/jquery/src

 src/Resources/public/vendor/jquery.scrollTo/demo
 src/Resources/public/vendor/jquery.scrollTo/tests
-

Files for master
----------------

diff --git a/.github/workflows/frontend.yaml b/.github/workflows/frontend.yaml
index d0fbdc3ae..f0b48a3a3 100644
--- a/.github/workflows/frontend.yaml
+++ b/.github/workflows/frontend.yaml
@@ -3,7 +3,6 @@ name: Frontend
 on:
     push:
         branches:
-            - 3.x
             - master
         paths:
             - assets/**
diff --git a/.gitignore b/.gitignore
index c3c28de3f..99700e9c5 100644
--- a/.gitignore
+++ b/.gitignore
@@ -11,62 +11,3 @@ phpunit.xml
 phpstan.neon
 /.phpunit.result.cache
 /docs/_build/
-
-# clean up some non bower ready package
-src/Resources/public/vendor/admin-lte/bootstrap
-src/Resources/public/vendor/admin-lte/build
-src/Resources/public/vendor/admin-lte/dist/img
-!src/Resources/public/vendor/admin-lte/dist/img/boxed-bg.jpg # Only image called in AdminLTE.css
-src/Resources/public/vendor/admin-lte/dist/js/app.js
-src/Resources/public/vendor/admin-lte/dist/js/demo.js
-src/Resources/public/vendor/admin-lte/dist/js/pages
-src/Resources/public/vendor/admin-lte/documentation
-src/Resources/public/vendor/admin-lte/pages
-src/Resources/public/vendor/admin-lte/plugins
-src/Resources/public/vendor/admin-lte/*.html
-src/Resources/public/vendor/admin-lte/Gruntfile.js
-
-src/Resources/public/vendor/iCheck/skins/all.css
-src/Resources/public/vendor/iCheck/skins/*/*
-!src/Resources/public/vendor/iCheck/skins/square/blue.css
-!src/Resources/public/vendor/iCheck/skins/square/blue.png
-!src/Resources/public/vendor/iCheck/skins/square/blue@2x.png
-src/Resources/public/vendor/iCheck/icheck.js
-
-src/Resources/public/vendor/slimScroll/examples
-src/Resources/public/vendor/slimScroll/jquery.slimscroll.js
-
-src/Resources/public/vendor/bootstrap/grunt
-src/Resources/public/vendor/bootstrap/less
-src/Resources/public/vendor/bootstrap/test-infra
-src/Resources/public/vendor/bootstrap/Gruntfile.js
-
-src/Resources/public/vendor/jqueryui/themes/black-tie
-src/Resources/public/vendor/jqueryui/themes/blitzer
-src/Resources/public/vendor/jqueryui/themes/cupertino
-src/Resources/public/vendor/jqueryui/themes/dark-hive
-src/Resources/public/vendor/jqueryui/themes/dot-luv
-src/Resources/public/vendor/jqueryui/themes/eggplant
-src/Resources/public/vendor/jqueryui/themes/excite-bike
-src/Resources/public/vendor/jqueryui/themes/hot-sneaks
-src/Resources/public/vendor/jqueryui/themes/humanity
-src/Resources/public/vendor/jqueryui/themes/le-frog
-src/Resources/public/vendor/jqueryui/themes/mint-choc
-src/Resources/public/vendor/jqueryui/themes/overcast
-src/Resources/public/vendor/jqueryui/themes/pepper-grinder
-src/Resources/public/vendor/jqueryui/themes/redmond
-src/Resources/public/vendor/jqueryui/themes/smoothness
-src/Resources/public/vendor/jqueryui/themes/south-street
-src/Resources/public/vendor/jqueryui/themes/start
-src/Resources/public/vendor/jqueryui/themes/sunny
-src/Resources/public/vendor/jqueryui/themes/swanky-purse
-src/Resources/public/vendor/jqueryui/themes/trontastic
-src/Resources/public/vendor/jqueryui/themes/ui-darkness
-src/Resources/public/vendor/jqueryui/themes/ui-lightness
-src/Resources/public/vendor/jqueryui/themes/vader
-
-src/Resources/public/vendor/jquery/src
-
-src/Resources/public/vendor/jquery.scrollTo/demo
-src/Resources/public/vendor/jquery.scrollTo/tests
-
```

jordisala1991:
Few things to resolve:

* There are tests and php-cs-fixer to fix.
* Maybe the `assets` option should be called `frontend`
* Maybe the json files should be converted to js, to be able to put de: `DO NOT TOUCH this file` (json does not have comments)

OskarStark:
Please also adjust `DispatchBranchesProtectionCommand` to make this build required

jordisala1991:
Related to: https://github.com/sonata-project/SonataAdminBundle/pull/7139

src/Command/Dispatcher/DispatchFilesCommand.php
```diff
@@ -295,6 +295,28 @@ private function deleteNotNeededFilesAndDirs(Project $project, Branch $branch, s
...
+        if (!$branch->hasFrontend()) {
+            $filepaths = [
+                '.babelrc.js',
+                '.eslintrc.js',
+                '.stylelintrc.js',
+                'postcss.config.js',
+                '.github/workflows/frontend.yaml',
+            ];
+
+            foreach ($filepaths as $filepath) {
```
OskarStark, line 307: Shall we extract a method to remove a file by path?

jordisala1991: I think it should be better now

src/Config/ProjectsConfiguration.php
```diff
@@ -52,6 +52,8 @@ public function getConfigTreeBuilder(): TreeBuilder
                                         ->arrayNode('tools')->prototype('scalar')->defaultValue([])->end()->end()
                                         ->arrayNode('php_extensions')->prototype('scalar')->defaultValue([])->end()->end()
                                         ->scalarNode('target_php')->defaultNull()->end()
+                                        ->scalarNode('custom_gitignore_part')->defaultNull()->end()
```
VincentLanglet, line 55: Some option like panther, phpstan, psalm are on the global level, and some options like php are in the branches level. Some option like custom_gitignore_part are now on both.

It seems weird to have some github related workflow `phpstan` on the global level but not `frontend`.
I understand why you did this but we should have the same strategy for every option IMHO.

Should we 
1) Move the all options to the branches level ? It will be more verbose (with duplicate values) but more configurable.
2) Duplicate all the options on the branches level and the global level ? When it's the same value for every branches (for instance php version) it will avoid duplicates lines.  
3) Wait for the 4.0 release to introduce the frontend option (directly on the global level ?)

Also if I use `custom_gitignore_part` on the branch level I would have expect to override the project level.
So I would have not expect the code
```
{% if branch.customGitignorePart is not empty %}
{%- elseif project.customGitignorePart is not empty -%}
```
But only a check on `branch.customGitignorePart` with a getter returning `project.customGitignorePart` by default if the branch level option is not set.

jordisala1991: IMO we should introduce options as we need them, to avoid implementing things just for having them. Remember this code is not intended to be public used, just internal tool for sonata.

And about what happens if you define at branch level and at root level, I think it should take precedence the branch over the project configuration, I thought I have done that part right, the if will check first for branch config, and then for project config.

VincentLanglet: Oh yes indeed, it an `elseif`.

jordisala1991:
Can we merge this one?

OskarStark (approved):

OskarStark:
I merged and deployed, can you please keep an eye on the DevKit PRs? Thanks

PR #1507, opened 2021-05-01
Allow to select the branch to release

VincentLanglet:
I used it for https://github.com/sonata-project/SonataBlockBundle/pull/879

image

### DIFF
--- a/config/packages/github_api.yaml
+++ b/config/packages/github_api.yaml
@@ -1,4 +1,10 @@
 services:
+    _defaults:
+        # The Github\Client is only calling authenticate/addCache when public (but I don't know why)
+        public: true
+        autowire: true
+        autoconfigure: true
+
     Redis:
         class: Redis
         calls:
--- a/config/packages/github_api.yaml
+++ b/config/packages/github_api.yaml
@@ -1,10 +1,4 @@
 services:
-    _defaults:
-        # The Github\Client is only calling authenticate/addCache when public (but I don't know why)
-        public: true
-        autowire: true
-        autoconfigure: true
-
     Redis:
         class: Redis
         calls:
--- a/src/Action/DetermineNextRelease.php
+++ b/src/Action/DetermineNextRelease.php
@@ -56,8 +56,9 @@ final class DetermineNextRelease
         try {
             $currentRelease = $this->releases->branchLatest($repository, $branch);
         } catch (LatestReleaseNotFound $e) {
-            throw CannotDetermineNextRelease::forProject(
+            throw CannotDetermineNextRelease::forBranch(
                 $project,
+                $branch,
                 $e
             );
         }
@@ -68,8 +69,9 @@ final class DetermineNextRelease
         );
 
         if ([] === $pullRequests) {
-            throw NoPullRequestsMergedSinceLastRelease::forProject(
+            throw NoPullRequestsMergedSinceLastRelease::forBranch(
                 $project,
+                $branch,
                 $currentRelease->publishedAt()
             );
         }
--- a/src/Action/DetermineNextRelease.php
+++ b/src/Action/DetermineNextRelease.php
@@ -54,7 +54,7 @@ final class DetermineNextRelease
         $repository = $project->repository();
 
         try {
-            $currentRelease = $this->releases->branchLatest($repository, $branch);
+            $currentRelease = $this->releases->latestForBranch($repository, $branch);
         } catch (LatestReleaseNotFound $e) {
             throw CannotDetermineNextRelease::forBranch(
                 $project,
--- a/src/Action/DetermineNextRelease.php
+++ b/src/Action/DetermineNextRelease.php
@@ -49,10 +49,9 @@ final class DetermineNextRelease
         $this->pullRequests = $pullRequests;
     }
 
-    public function __invoke(Project $project, ?Branch $branch = null): NextRelease
+    public function __invoke(Project $project, Branch $branch): NextRelease
     {
         $repository = $project->repository();
-        $branch = $branch ?? $project->stableBranch() ?? $project->unstableBranch();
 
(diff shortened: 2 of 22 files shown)


VincentLanglet:
I made the code BC for the controller which display the next release.
image

But should we update them in order to display all the branches ?
This way we won't miss a potential release (for instance for the block-bundle@3.x).

WDYT @sonata-project/contributors ?

VincentLanglet:
I need to make more test for release on master branch. But I'm currently getting
```
 You have reached GitHub hourly limit! Actual limit is: 60
```

jordisala1991:
I like the idea, and it is probably a good solution, but we have a problem with the github api calls. One question, does packagist expose an api for releases?

VincentLanglet:
I don't know why but I'm getting
```
[WARNING] Release is not needed
```
When trying to run the command with `admin-bundle` and `master` branch.

cc @OskarStark

VincentLanglet:
> I like the idea, and it is probably a good solution, but we have a problem with the github api calls. One question, does packagist expose an api for releases?

It's not supposed to make a lot more calls than before.
- When running `bin/console release` it's supposed to do the same calls.
- On the controller, it will make more calls for project with multiple stable branch, I think `block-bundle` is the only one. So it's just like having one project more to check for release.

I'm not sure packagist would be better since we still have a lot of github checks to do. We should look for caching calls or having more call to github per hour...

jordisala1991:
we should have 5000 calls per hour: https://docs.github.com/es/rest/overview/resources-in-the-rest-api#rate-limiting

something is wrong here

jordisala1991:
Did you add your github token on .env?

VincentLanglet:
> Did you add your github token on .env?

Yes. But when I dump the authenticate method it seems like it's not called.
It's like the config
```
    Github\Client:
        arguments:
            - '@Github\HttpClient\Builder'
        calls:
            - ['authenticate', ['%env(GITHUB_OAUTH_TOKEN)%', null, 'access_token_header']]
            - ['addCache', ['@Cache\Adapter\Redis\RedisCachePool']]
```
is not used

VincentLanglet:
Yes, if I call manually `$this->github->authenticate` it works.

VincentLanglet:
This also explain why the cache is not used on the website.

Now, how to fix this ?

VincentLanglet:
Adding
```
    _defaults:
        # The Github\Client is only calling authenticate/addCache when public (but I don't know why)
        public: true
        autowire: true
        autoconfigure: true
```
Does call `authenticate`

I don't know why it need to be public...

config/packages/github_api.yaml
```diff
@@ -1,4 +1,10 @@
 services:
+    _defaults:
+        # The Github\Client is only calling authenticate/addCache when public (but I don't know why)
```
OskarStark, line 3: Nice catch, can you open an issue on the repo and ask?

VincentLanglet: Should it be on the Github\Client repo or on the symfony repo ?

franmomu: I'm curious why that would happen, just making `Github\Client` public fixes the issue or should all services defined there be public?

I've never used the commands.

src/Action/DetermineNextRelease.php
```diff
@@ -49,24 +49,16 @@ public function __construct(
...
-        } catch (NoBranchesAvailable $e) {
-            throw CannotDetermineNextRelease::forProject(
-                $project,
-                $e
-            );
-        }
-
-        try {
-            $currentRelease = $this->releases->latest($repository);
+            $currentRelease = $this->releases->branchLatest($repository, $branch);
```
franmomu, line 57:
```suggestion
            $currentRelease = $this->releases->latestBranch($repository, $branch);
```
(I think)

jordisala1991: The method fetch latest release from given branch, maybe should simply be latestRelease, or more verbose latestBranchRelease

VincentLanglet: There is a `latest` method. It would be weird to add `Release` to this method and not the `latest` one.

Here I wanted to say it's the branch's latest release. So branchLatest.
Maybe latestForBranch is more clear ?

config/packages/github_api.yaml
VincentLanglet: I just discovered that I don't need to add
```
    _defaults:
        # The Github\Client is only calling authenticate/addCache when public (but I don't know why)
        public: true
        autowire: true
        autoconfigure: true
```

If I remove the call 
```
- ['addCache', ['@Cache\Adapter\Redis\RedisCachePool']]
```

Maybe it's related to the failure I get
```
                                                                                                                        
 [WARNING] Some commands could not be registered:                                                                       
                                                                                                                        

In getClientService.php line 53:
                      
  Connection refused  
```

The failure of one calls is denying every calls ? how to solve this ?

VincentLanglet: @sonata-project/contributors Any idea why the `addCache` call is disabling the github authentication ?

@OskarStark should we remove it ?

jordisala1991: https://github.com/KnpLabs/php-github-api/issues/947

Maybe this helps?

VincentLanglet: Ok I found the reason.

When I correctly configure `redis` the `authenticate` call is made.
So it was related to the error
```
 [WARNING] Some commands could not be registered:                                                                       
                                                                                                                        

In getClientService.php line 53:
                      
  Connection refused  
```

Is there a way to add the `addCache` call only if redis server is started ? ^^'

VincentLanglet:
I think this is ready to be merged :)

OskarStark (approved):

PR #1546, opened 2021-05-25
Add compilation check for frontend ci

VincentLanglet:
POC: https://github.com/sonata-project/SonataAdminBundle/pull/7221

### DIFF
--- a/templates/project/.github/workflows/frontend.yaml.twig
+++ b/templates/project/.github/workflows/frontend.yaml.twig
@@ -38,3 +38,6 @@ jobs:
 
             - name: Run Webpack Encore
               run: yarn encore production
+
+            - name: Check if the compilation is up to date
+              run: git diff --no-patch --exit-code src/Resources/public


templates/project/.github/workflows/frontend.yaml.twig
```diff
@@ -38,3 +38,6 @@ jobs:
 
             - name: Run Webpack Encore
               run: yarn encore production
+
+            - name: Check if the compilation is up to date
+              run: git diff --no-patch --exit-code src/Resources/public/app.js src/Resources/public/app.css
```
core23, line 43: We don't need this for every project

core23 (changes requested):

templates/project/.github/workflows/frontend.yaml.twig
jordisala1991: Is there a way to check for everything in web/Resources/public? Every project with frontend enabled should check for that directory (atm only admin bundle)

VincentLanglet: @core23 We need this for every project with the frontend github actions.

@jordisala1991 I made the change

core23 (approved):

jordisala1991 (approved):

PR #1632, opened 2021-08-13
Add php_unit_test_case_static_method_calls

franmomu:
This will change all the PHPUnit dynamic calls to static calls and help with the strict rules phpstan plugin

Or do we want to do this when the time comes for each package?

Issue #1559, opened 2021-06-06
Use both phpunit/phpunit and simple-phpunit.


VincentLanglet:
See https://github.com/symfony/test-pack/pull/10
https://github.com/symfony/recipes/pull/906

Symfony still started to add phpunit/phpunit to the simple-phpunit recipe.

WDYT about adding the dependency @sonata-project/contributors ?
It will solve the need of the bootstrap file for static analysis: https://github.com/sonata-project/SonataAdminBundle/blob/master/phpstan.neon.dist#L7

jordisala1991:
Solved

PR #1634, opened 2021-08-15
Upgrade to php-cs-fixer 3

franmomu:
this should fix https://github.com/sonata-project/dev-kit/pull/1632#pullrequestreview-730162680

POC here: https://github.com/sonata-project/SonataAdminBundle/pull/7434

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -17,7 +17,7 @@ jobs:
               uses: actions/checkout@v2.3.4
 
             - name: Run PHP-CS-Fixer
-              uses: docker://oskarstark/php-cs-fixer-ga:2.18.1
+              uses: docker://oskarstark/php-cs-fixer-ga:3.0.0
               with:
                   args: --ansi --verbose --diff --dry-run
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/.php_cs.cache
+/.php-cs-fixer.cache
 /.env
 /.bundle
 /docs/_build/
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -1,6 +1,6 @@
 <?php
 
-$config = include 'templates/project/.php_cs.dist';
+$config = include 'templates/project/.php-cs-fixer.dist.php';
 
 $finder->in(__DIR__);
 
--- a/templates/project/.github/workflows/lint.yaml.twig
+++ b/templates/project/.github/workflows/lint.yaml.twig
@@ -23,7 +23,7 @@ jobs:
               uses: actions/checkout@v2
 
             - name: Run PHP-CS-Fixer
-              uses: docker://oskarstark/php-cs-fixer-ga:2.18.5
+              uses: docker://oskarstark/php-cs-fixer-ga:3.0.0
               with:
                   args: --ansi --verbose --diff --dry-run
 
--- a/templates/project/.gitignore.twig
+++ b/templates/project/.gitignore.twig
@@ -5,7 +5,7 @@
 /build/
 /node_modules/
 /vendor/
-.php_cs.cache
+.php-cs-fixer.cache
 composer.lock
 phpunit.xml
 phpstan.neon
--- a/templates/project/.php-cs-fixer.dist.php
+++ b/templates/project/.php-cs-fixer.dist.php
@@ -16,8 +16,6 @@ file that was distributed with this source code.
 HEADER;
 
 $rules = [
-    '@PHP56Migration' => true,
-    '@PHP56Migration:risky' => true,
     '@PHP70Migration' => true,
     '@PHP70Migration:risky' => true,
     '@PHP71Migration' => true,
@@ -69,9 +67,12 @@ $finder = PhpCsFixer\Finder::create()
     ->exclude('var')
 ;
 
-return PhpCsFixer\Config::create()
+$config = new PhpCsFixer\Config();
+$config
     ->setFinder($finder)
     ->setRiskyAllowed(true)
     ->setRules($rules)
     ->setUsingCache(true)
 ;
+
+return $config;


jordisala1991 (approved):

VincentLanglet:
When running `php-cs-fixer` on a project
https://github.com/sonata-project/SonataAdminBundle/blob/4.x/Makefile#L46

Am I wrong or are we using the global installation of php-cs-fixer ?
- It require to update regularly the github actions to bump the versions
- It require for the developer to follow the global installation of php-cs-fixer
- It require for the developer to bump the version regularly of the global installation

Shouldn't we use the same strategy than phpstan/psalm by requiring-dev it in the composer.json and using this version for both the Makefile and the github action ?

VincentLanglet:
> When running `php-cs-fixer` on a project
> https://github.com/sonata-project/SonataAdminBundle/blob/4.x/Makefile#L46
> 
> Am I wrong or are we using the global installation of php-cs-fixer ?
> 
> * It require to update regularly the github actions to bump the versions
> * It require for the developer to follow the global installation of php-cs-fixer
> * It require for the developer to bump the version regularly of the global installation
> 
> Shouldn't we use the same strategy than phpstan/psalm by requiring-dev it in the composer.json and using this version for both the Makefile and the github action ?

For instance, the version should already be updated to 3.0.2 now.

jordisala1991:
+1 to that, maybe it is a good idea to add phpcs fixer and phpunit at the same time, so we have only to touch one time the composer.json

VincentLanglet:
> +1 to that, maybe it is a good idea to add phpcs fixer and phpunit at the same time, so we have only to touch one time the composer.json

I just tried to add phpunit on a project and I see one issue
- The phpunit version is decided by dev-kit https://github.com/sonata-project/dev-kit/blob/master/templates/project/phpunit.xml.dist.twig#L38
So if we add the dependency to the composer.json, it will be handled at two place in the dev-kit and in the bundle.
Is there a way to avoid this ?

jordisala1991:
Yes, remove that line on dev kit , since we wont be installing simple-phpunit

VincentLanglet:
> since we wont be installing simple-phpunit

Not sure about this point, we still need feature like
```
<env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
```

The issue https://github.com/sonata-project/dev-kit/issues/1559 was talking about having both phpunit and simple-phpunit in the dependencies (since it's now kinda recommended). But if it's too complicated, we can just stick with the way we're doing with only simple-phpunit.

jordisala1991:
Read the last point on this : https://symfony.com/doc/current/components/phpunit_bridge.html#usage 

You have to register a listener and thats all

VincentLanglet:
> Read the last point on this : https://symfony.com/doc/current/components/phpunit_bridge.html#usage
> 
> You have to register a listener and thats all

But, in order to register the listener, the bridge still need to be installed.

jordisala1991:
> > Read the last point on this : https://symfony.com/doc/current/components/phpunit_bridge.html#usage
> 
> > 
> 
> > You have to register a listener and thats all
> 
> 
> 
> But, in order to register the listener, the bridge still need to be installed.

Yes, both need to be installed but there is no need to invoke simple-phpunit unless you need to test with multiple phpunit version, which we dont (atm)

franmomu:
I'm fine adding them to the packages, maybe it could be interesting to have a way in `dev-kit` to create PRs as a result of a `composer` command or something like that to not have to create manually all the PRs.

Anyway, even if we add it individually, this PR would ease the transition upgrading the configuration for php-cs-fixer.

VincentLanglet:
> Yes, both need to be installed but there is no need to invoke simple-phpunit unless you need to test with multiple phpunit version, which we dont (atm)

If you know exactly what to do I'm not against a POC PR, then I'll replicate it.

> I'm fine adding them to the packages, maybe it could be interesting to have a way in dev-kit to create PRs as a result of a composer command or something like that to not have to create manually all the PRs.

I don't know how hard it would be to do this, but sure it could be interesting.
Until then, I'll survive doing those PR.

> Anyway, even if we add it individually, this PR would ease the transition upgrading the configuration for php-cs-fixer.

Sure I didn't want to be a blocker

.github/workflows/lint.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
               uses: actions/checkout@v2.3.4
 
             - name: Run PHP-CS-Fixer
-              uses: docker://oskarstark/php-cs-fixer-ga:2.18.1
+              uses: docker://oskarstark/php-cs-fixer-ga:3.0.0
```
VincentLanglet, line 20:
```suggestion
              uses: docker://oskarstark/php-cs-fixer-ga:3.0.2
```

VincentLanglet (approved):

VincentLanglet:
I don't know why it fail, it's supposed to be released: https://github.com/OskarStark/php-cs-fixer-ga/releases/tag/3.0.2

franmomu:
Looks like the tag in docker https://hub.docker.com/r/oskarstark/php-cs-fixer-ga/tags?page=1&ordering=last_updated is missing, ping @OskarStark

OskarStark:
> Looks like the tag in docker https://hub.docker.com/r/oskarstark/php-cs-fixer-ga/tags?page=1&ordering=last_updated is missing, ping @OskarStark

Yes it is, because docker stopped auto builds feature for non pro (paid) accounts and I currently don't have time to migrate to GitHub action and I don't want to pay 60$ a year 😞 

Anyone open for a PR to build and push a tag on `oskarstark/php-cs-fixer-ga` ? I will provide the credentials as SECRETS and add you as a collaborator, that the workflow runs in the PR

franmomu:
> I don't want to pay 60$ a year 😞

That's more than fair.

> Anyone open for a PR to build and push a tag on `oskarstark/php-cs-fixer-ga` ? I will provide the credentials as SECRETS and add you as a collaborator, that the workflow runs in the PR

I've only used docker just as a provider of services, never tried building or tagging.

I've changed again to use php-cs-fixer version to 3.0.0, since the idea of this PR is to make the upgrade of configuration and files automatically.

OskarStark:
@franmomu I now have a paid plan, you can create a PR for 3.0.2 now, cheers

PR #1643, opened 2021-08-26
Split test and coverage command

VincentLanglet:
I have pcov locally, so the test command try to generate coverage, and the memory consumption explode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,12 +64,12 @@ cs-fix-xml:
 .PHONY: cs-fix-xml
 
 test:
-ifeq ($(shell php --modules|grep --quiet pcov;echo $$?), 0)
-	vendor/bin/simple-phpunit -c phpunit.xml.dist --coverage-clover build/logs/clover.xml
-else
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -64,13 +64,13 @@ cs-fix-xml:
...
 
 test:
-ifeq ($(shell php --modules|grep --quiet pcov;echo $$?), 0)
-	vendor/bin/simple-phpunit -c phpunit.xml.dist --coverage-clover build/logs/clover.xml
-else
 	vendor/bin/simple-phpunit -c phpunit.xml.dist
-endif
 .PHONY: test
 
+coverage:
```
OskarStark, line 70: test-with-coverage 

?

OskarStark (approved):

Makefile
VincentLanglet: kinda long to write no ?

I also was wondering if I should say something about clover.
Because we might want to run coverage-html locally instead of clover in order to see what is the uncovered code ?

jordisala1991 (approved):

PR #1106, opened 2020-09-24
[RFC] Allow core members to be advertised for sponsoring

OskarStark:
I checked all core members, but only @core23 and me are registered for GithubSponsors.

Are some of you using another platform such as:
CleanShot 2020-09-24 at 08 45 08

WDYT?

PR #1646, opened 2021-08-28
Dynamically generate FUNDING

VincentLanglet:
Following the strategy of https://github.com/sonata-project/dev-kit/pull/1106,  I wanted to add greg0ire to the sponsor list.
But only 4 members are allowed.
So I added some logic: for a repository, it list the 4 most active members. Seems like a fair rule to me.


wbloszyk (approved):

core23:
> But only 4 members are allowed.

This is a [GitHub limit](https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/displaying-a-sponsor-button-in-your-repository#about-funding-files).

Sounds fair to list the most active developers as long as this limit exist. Should we list *all* developers in our `composer.json` file? There is [no limit](https://getcomposer.org/doc/04-schema.md#funding), but it would require some more work to update the file on the fly.

VincentLanglet:
> Should we list _all_ developers in our `composer.json` file? There is [no limit](https://getcomposer.org/doc/04-schema.md#funding), but it would require some more work to update the file on the fly.

Currently there is no modification of the `composer.json` by the dev-kit.
This could be an idea but would require some work indeed.

I'm fine with the current 
```
{
    "name": "Sonata Community",
    "homepage": "https://github.com/sonata-project/SonataAdminBundle/contributors"
}
```

core23 (approved):

PR #1670, opened 2021-09-14
Try with Symfony 5.4

jordisala1991:
We need to add test for Symfony 5.4 and start fixing deprecations. Why? Because Symfony will add a lot of types on 6.0 and we can detect those on Symfony 5.4 already.

They said we can report the types that can't be added without BC break post them here: https://github.com/symfony/symfony/issues/43021

There is also a tool that can be used to automatically add types and detect incompatibilities: https://wouterj.nl/2021/09/symfony-6-native-typing

IMO we should slow down a bit and ensure we are fully compatible with Sf 5.4 and 6.0 because we have a good change to be compatible right away and if not maybe symfony guys can help us. If we delay this a lot, we would end up in the same situation like with Symfony 5.

wdyt?

### DIFF
--- a/templates/project/.github/workflows/test.yaml.twig
+++ b/templates/project/.github/workflows/test.yaml.twig
@@ -99,6 +99,9 @@ jobs:
               if: matrix.variant != 'normal' && !startsWith(matrix.variant, 'symfony/symfony')
               {% verbatim %}run: composer require ${{ matrix.variant }} --no-update{% endverbatim %}
 
+            - name: Allow unstable dependencies
+              run: composer config minimum-stability dev
+
 {# Remove when removing 3.x for the managed branches of block-bundle #}
 {% if project.repository.name == 'SonataBlockBundle' and branch.name == '3.x' %}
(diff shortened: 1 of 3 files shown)


templates/project/.github/workflows/test.yaml.twig
```diff
@@ -99,6 +99,9 @@ jobs:
               if: matrix.variant != 'normal' && !startsWith(matrix.variant, 'symfony/symfony')
               {% verbatim %}run: composer require ${{ matrix.variant }} --no-update{% endverbatim %}
 
+            - name: Allow unstable dependencies
+              run: composer config minimum-stability dev
```
franmomu, line 103: Should this be only executed when Symfony `5.4`?

jordisala1991: It doesnt matter, since we are using prefer-stable , when there is a stable release it will be used, but for sf 5.4 there is no stable release so it will use the dev stability

franmomu (approved):

VincentLanglet (approved):

PR #1683, opened 2021-09-18
Stop excluding readme

VincentLanglet:
I don't think we need to exclude readme for those projects.
And it would be great to have no readme excluded.

### DIFF
--- a/config/projects.yaml
+++ b/config/projects.yaml
@@ -206,8 +206,6 @@ datagrid-bundle:
 doctrine-extensions:
     phpstan: true
     psalm: true
-    excluded_files:
-        - README.md
     has_documentation: false
     branches:
         2.x:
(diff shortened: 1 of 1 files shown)


config/projects.yaml
```diff
@@ -360,8 +358,6 @@ form-extensions:
                 symfony/symfony: ['4.4.*', '5.3.*', '5.4.*', '6.0.*']
 
 google-authenticator:
-    excluded_files:
-        - README.md
```
jordisala1991, line 364: this project has no docs, and I think it does have some information in the readme

VincentLanglet: indeed

jordisala1991 (approved):

OskarStark:
Please double check the dev kit PR afterwards 👍🏻

OskarStark (approved):

PR #1724, opened 2021-10-17
Use phpunit from installed vendors

jordisala1991:
Proof: https://github.com/sonata-project/SonataAdminBundle/pull/7540

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -21,13 +21,9 @@ jobs:
               with:
                   php-version: 8.0
                   coverage: none
-                  tools: composer:v2
-
-            - name: Install Composer dependencies
-              uses: ramsey/composer-install@v1
-              with:
-                  dependency-versions: highest
-                  composer-options: --prefer-dist --prefer-stable
+                  tools: composer:v2, php-cs-fixer:3
+              env:
+                  COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
             - name: Lint PHP files
               run: make lint-php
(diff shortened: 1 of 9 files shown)


jordisala1991:
For me I would do it only for phpunit for the moment (we also have composer-normalizer usage and we do not install on composer.json)

But @VincentLanglet requested it for both tools. I can't make both of you happy it seems. What should I do?

core23:
> But @VincentLanglet requested it for both tools. I can't make both of you happy it seems. What should I do?

Install phpunit as a dev dependency and all other tools via separate composer config?

jordisala1991:
With this PR we have a few good things solved:

* PHPUnit installed on each project (as Symfony recommends and it solves the build problem)
* php-cs-fixer updated to last v3 automatically (before it depended on the @OskarStark action)
* composer normalizer updated to use the same strategy as php-cs-fixer (also the action is kind of deprecated)

Whats missing?

* A specific composer.json for tools (IMO it can be done later if we need to)
* The ability to run php-cs-fixer and composer-normalize on each contributor computer (that needs a specific composer.json for tools or add them to our composer.json which @core23 do not agree)

The missing points can be solved later IMHO

VincentLanglet:
> But @VincentLanglet requested it for both tools. I can't make both of you happy it seems. What should I do?

I won't make phpcsfixer as a blocking point if @core23 is not happy with it.

My issue is that I am not able to run phpcsfixer on sonata projects because it require it globally and I cannot change the version I have globally.

But it could be great to find a solution later to
> The ability to run php-cs-fixer and composer-normalize on each contributor computer (that needs a specific composer.json for tools or add them to our composer.json which @core23 do not agree)

Still, thanks for the work @jordisala1991 :)

Issue #1433, opened 2021-03-01
Support for PHP 8.0


wbloszyk:
Getting from sandbox(composer why-not php 8.0):
```
sonata-project/sandbox                dev-master   requires  php (^7.4)            
doctrine/orm                          9999999-dev  requires  php (^7.1)            
doctrine/reflection                   1.3.x-dev    requires  php (^7.1)            
friendsofsymfony/comment-bundle       2.x-dev      requires  php (^5.5.9|~7.0)     
friendsofsymfony/user-bundle          2.x-dev      requires  php (^5.5.9 || ^7.0)  
knplabs/knp-markdown-bundle           1.8.1        requires  php (^7.1.3)          
kriswallsmith/buzz                    v0.16.0      requires  php (^5.4 || ^7.0)    
sonata-project/cache                  1.x-dev      requires  php (^7.1)            
sonata-project/cache-bundle           2.x-dev      requires  php (^5.6 || ^7.0)    
sonata-project/classification-bundle  3.13.2       requires  php (^7.2)            
sonata-project/comment-bundle         9999999-dev  requires  php (^7.3)            
sonata-project/easy-extends-bundle    9999999-dev  requires  php (^7.1)            
sonata-project/ecommerce              9999999-dev  requires  php (^7.3)            
sonata-project/formatter-bundle       9999999-dev  requires  php (^7.3)            
sonata-project/media-bundle           3.30.0       requires  php (^7.3)            
sonata-project/news-bundle            9999999-dev  requires  php (^7.3)            
sonata-project/notification-bundle    3.11.0       requires  php (^7.3)            
sonata-project/page-bundle            9999999-dev  requires  php (^7.2)            
sonata-project/timeline-bundle        9999999-dev  requires  php (^7.3)            
sonata-project/user-bundle            9999999-dev  requires  php (^7.3)            
stephpy/timeline-bundle               3.1.1        requires  php (^7.1)     
```
### TODO:

- [x] sonata-project/cache
- [x] sonata-project/cache-bundle
- [x] sonata-project/classification-bundle 
- [ ] sonata-project/comment-bundle 
- [ ] sonata-project/ecommerce
- [x] sonata-project/formatter-bundle
- [x] sonata-project/media-bundle
- [ ] sonata-project/news-bundle
- [x] sonata-project/page-bundle
- [ ] sonata-project/timeline-bundle
- [x] sonata-project/user-bundle

wbloszyk:
[...]

PR #1728, opened 2021-10-17
admin-bundle: Add target_php 8.0

franmomu:
Adding a `target_php` will use that php version for the jobs instead of the last one, otherwise when allowing `8.1`, it will be use for every job: https://github.com/sonata-project/SonataAdminBundle/pull/7543/files

The removed part should be fine since https://github.com/sonata-project/dev-kit/issues/1433 is solved.

### DIFF
--- a/config/projects.yaml
+++ b/config/projects.yaml
@@ -128,9 +128,9 @@ cache:
     has_documentation: false
     branches:
         3.x:
-            php: ['7.3', '7.4']
+            php: ['7.3', '7.4', '8.0']
         2.x:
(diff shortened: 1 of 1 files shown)


franmomu:
I've added to `cache` and `google-authenticator` which already supports 8.0 in their `composer.json`.

The packages remaining are (all deprecated I think):
- sandbox
- timeline-bundle
- ecommerce
- doctrine-phpcr-admin-bundle
- dashboard-bundle:
- comment-bundle
- classification-media-bundle
- article-bundle
- admin-search-bundle

PR #1751, opened 2021-11-02
Allow @psalm-suppress in phpdoc outside structural elements

jordisala1991:
Otherwise it gets converted to normal php comments and not parsed by Psalm.


VincentLanglet (approved):
I didnt see my feature was released. Happy to finally use it.

PR #1755, opened 2021-11-05
Make phpstan and psalm true by default

jordisala1991:
This way we can easily know which projects are missing. This is mostly
because now 90% of the projects have both tools and only a few are
missing.


jordisala1991:
Missing packages on projects.yml:

* cache
* cache-bundle
* datagrid-bundle
* intl-bundle
* page-bundle
* formatter-bundle (only psalm)
* sandbox (only psalm)

greg0ire (approved):

VincentLanglet:
Can you try to add phpstan on datagrid ? https://github.com/sonata-project/SonataDatagridBundle/blob/3.x/phpstan.neon.dist

VincentLanglet (approved):

core23 (approved):

jordisala1991:
> Can you try to add phpstan on datagrid ? https://github.com/sonata-project/SonataDatagridBundle/blob/3.x/phpstan.neon.dist

Should be done on another Pr, It might require some changes on datagrid

PR #1762, opened 2021-11-07
Dispatch on github actions

jordisala1991:


core23 (approved):

.github/workflows/dispatch.yaml
```diff
@@ -0,0 +1,44 @@
+name: Dispatch
+
+on:
+    push:
+        branches:
+            - master
+    pull_request:
```
core23, line 7: Do we need this for pull requests too?

jordisala1991: Wait, Im just testing 😅 

My idea was to dispatch on pull request and `--apply` on pushes. So you could see what will be dispatched, right on the pull request.

jordisala1991: I would like to also add a repository_dispatch (not sure if that's the name) so we can run it right in the interface of the actions.

And of course we need the cron , so it runs daily.

jordisala1991: What I am not sure if we need the push (it could cause a lot of PRs on pushes here, for example if 2 Prs get merged one after another.

jordisala1991:
Sorry, we need to continue discussion here. To be able to access secrets I have to create the branch on the repository itself, not the fork. https://github.com/sonata-project/dev-kit/pull/1763

.github/workflows/dispatch.yaml
jordisala1991: Let's continue this on the new PR.

PR #1763, opened 2021-11-07
Dispatch on github actions

jordisala1991:


jordisala1991:
WDYT @sonata-project/contributors ?

The main idea about this PR is:
* We change files here often
* We have to remember to deploy each change to Symfony Cloud
* If we forget about it, changes don't get dispatched and we lose time
* Github actions offer cronjobs already, why don't we use that directly?

With this one we should be able to stop worrying about deploying each change to Symfony Cloud. I didn't migrate all crons because I thought it would be better starting with something, and if it works well then we can migrate the rest.

Symfony Cloud is still needed because we have a web interface with some nice tools. (But it could be handled via github actions too if we think about it. Not to be done here, but just food for thoughts).

jordisala1991:
Of course it can be also improved later to:

* Not only have cronjob but a button to manually dispatch files
* Have a special comment on PRs to trigger a dispatch to show changes (without --apply)

jordisala1991:
If we accept this change, the person who merges it should ensure to deploy one more time to Symfony Cloud to remove those crons.

VincentLanglet (approved):

OskarStark (approved):

VincentLanglet:
merged and deployed

PR #1790, opened 2021-11-28
Remove Licence from excluded files

VincentLanglet:
I asked beberlei, and he's ok with the change of the Licence of the EntityAuditBundle from GNU to MIT.

We should also have the agreement of main contributors.
https://github.com/sonata-project/EntityAuditBundle/graphs/contributors

@bendavies @andrewtch are you also ok with this change ?


bendavies:
that's fine

PR #1791, opened 2021-11-28
Dispatch files even if the branch is already created

jordisala1991:
Also:

* Remove annoying output for deleted files
* Add workflow dispatch for dispatch files and branches protection

### DIFF
--- a/.github/workflows/dispatch-files-branches.yaml
+++ b/.github/workflows/dispatch-files-branches.yaml
@@ -9,7 +9,7 @@ on:
         workflow_dispatch:
             inputs:
                 projects:
-                    description: 'Projects to dispatch files, separated by spaces (empty means all)'
+                    description: 'Projects to dispatch files and branches protection, separated by spaces (empty means all)'
                     default: ''
 
 env:
--- a/.github/workflows/dispatch-files-branches.yaml
+++ b/.github/workflows/dispatch-files-branches.yaml
@@ -9,7 +9,6 @@ on:
         inputs:
             projects:
                 description: 'Projects to dispatch files and branches protection, separated by spaces (empty means all)'
-                default: ''
 
 env:
     REQUIRED_PHP_EXTENSIONS: redis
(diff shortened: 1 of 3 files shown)


jordisala1991:
One is for the fix about some projects not getting the new changes (there are commits that couldn't be created as PR because of some rate limit on the api and dev-kit doesn't diff with the main branch)

The other is to be able to execute the action manually (with a projects parameter to only execute for some project if needed)

src/Command/Dispatcher/DispatchFilesCommand.php
```diff
@@ -291,7 +294,6 @@ private function deleteNotNeededFilesAndDirs(Project $project, Branch $branch, s
                 ->toString();
 
             if ($this->filesystem->exists($file)) {
-                $this->io->writeln(sprintf('Delete <info>/%s</info> file!', $fileToRemove));
```
jordisala1991, line 294: This output is annoying because everytime it says what is going to be removed, even if its already removed. IMO not needed. If something is about to be removed it will show on the diff

.github/workflows/dispatch-files-branches.yaml
```diff
@@ -5,6 +5,12 @@ on:
         # At minute 5 past every 4th hour
         # https://crontab.guru/#5_*/4_*_*_*
         - cron: '5 */4 * * *'
+    on:
+        workflow_dispatch:
+            inputs:
+                projects:
+                    description: 'Projects to dispatch files, separated by spaces (empty means all)'
```
jordisala1991, line 12:
```suggestion
                    description: 'Projects to dispatch files and branches protection, separated by spaces (empty means all)'
```

.github/workflows/dispatch-files-branches.yaml
```diff
@@ -5,6 +5,11 @@ on:
         # At minute 5 past every 4th hour
         # https://crontab.guru/#5_*/4_*_*_*
         - cron: '5 */4 * * *'
+    workflow_dispatch:
+        inputs:
+            projects:
+                description: 'Projects to dispatch files and branches protection, separated by spaces (empty means all)'
+                default: ''
```
jordisala1991, line 12:
```suggestion
```

VincentLanglet (approved):

PR #1793, opened 2021-11-28
Remove 8.1 to older release

jordisala1991:


jordisala1991:
4.x branch should be created on translation bundle @franmomu 
2.x is too old and not worth it to test with 8.1

PR #1802, opened 2021-12-02
Psalm plugin supports SF6

VincentLanglet:


jordisala1991:
We can probably remove the lines before too. Keep in mind dev kit will open a some PRs with broken builds.

VincentLanglet:
> We can probably remove the lines before too. Keep in mind dev kit will open a some PRs with broken builds.

This will be a nice way to track broken build and solve them

jordisala1991 (approved):

PR #1813, opened 2021-12-05
Only allow dev dependencies for SF6

VincentLanglet:
We should use stable dependencies for build as much as possible.


jordisala1991:
Yes, but still this should be merged: https://github.com/sonata-project/SonataAdminBundle/pull/7636

jordisala1991 (approved):

PR #1814, opened 2021-12-06
Allow doctrine test bundle to correctly isolate database tests

jordisala1991:


config/projects.yaml
```diff
@@ -398,6 +398,7 @@ twig-extensions:
                 symfony/symfony: ['4.4.*', '5.3.*', '5.4.*', '6.0.*']
 
 user-bundle:
+    doctrine_test: true
```
core23, line 401: Just an idea. What about defining a list of custom phpunit extensions instead of a boolean value.

templates/project/phpunit.xml.dist.twig
```diff
@@ -30,9 +30,14 @@ It's auto-generated by sonata-project/dev-kit package.
...
     </listeners>
 
-{% if project.usesPanther %}
+{% if project.usesPanther or project.usesDoctrineTest %}
     <extensions>
+{% if project.usesPanther %}
         <extension class="Symfony\Component\Panther\ServerExtension" />
+{% endif %}
+{% if project.usesDoctrineTest %}
+        <extension class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension" />
```
VincentLanglet, line 39: This extension is pretty useful, I'm kinda curious why we don't need it on DoctrineORM bundle

config/projects.yaml
VincentLanglet: This might be a better solution indeed to avoid too many config key.

jordisala1991: I will try to take a look at it.

templates/project/phpunit.xml.dist.twig
jordisala1991: Yes, To me it should be added to the projects where we use functional tests. Now we are not using it because we avoid somehow colisions between tests. On media bundle wasn't needed because of that, and classification the same. On user-bundle I can't test properly without it, because some properties are unique and it is a pain to deal with them on an non isolated environment.

config/projects.yaml
jordisala1991: Done.

templates/project/phpunit.xml.dist.twig
jordisala1991: Once we have a example setup, we can use it on more bundles if needed.

VincentLanglet (approved):

tests/Command/Dispatcher/DispatchFilesCommandTest.php
```diff
@@ -75,7 +75,7 @@ public function projectProvider(): iterable
   composer_version: '1'
   phpstan: true
   psalm: true
-  panther: true
+  phpunit_extensions: ['panther', 'doctrine_test']
```
core23, line 78: Would prefer something like this:

```suggestion
  phpunit_extensions: 
    - Symfony\Component\Panther\ServerExtension
    - DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension
```

core23 (approved):

tests/Command/Dispatcher/DispatchFilesCommandTest.php
jordisala1991: I dont think thats a good idea, the phpunit extensions (in case of panther) are not only used to enable the extension. Panther needs an env var and some gitignore files.

PR #1843, opened 2022-01-02
Add Symfony lint commands

jordisala1991:
Example: https://github.com/sonata-project/SonataAdminBundle/pull/7678


templates/project/bin/console.twig
```diff
@@ -0,0 +1,29 @@
...
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Sonata\{{ bundle_name }}\Tests\App\AppKernel;
```
jordisala1991, line 15: I need to be able to generate this bundle_name, do you know if we already have this information somehow in dev-kit? @VincentLanglet

jordisala1991:
We currently have this situation:

## Considered Bundles by dev-kit (The check is done by finding bundle in the name of the project)

- [x] admin-bundle
- [x] classification-bundle
- [x] doctrine-orm-admin-bundle
- [x] media-bundle
- [x] page-bundle
- [x] translation-bundle
- [x] user-bundle
- [ ] doctrine-mongodb-admin-bundle (from 4.x)
- [ ] block-bundle (from 4.x)
- [ ] seo-bundle
- [ ] entity-audit-bundle
- [ ] formatter-bundle
- [ ] intl-bundle
- [ ] cache-bundle (abandoned)
- [ ] news-bundle (abandoned)
- [ ] notification-bundle (abandoned)
- [ ] datagrid-bundle (abandoned)

## Others that have Symfony integration

- [x] doctrine-extensions (rename from Kernel to AppKernel)
- [x] twig-extensions
- [ ] exporter
- [ ] form-extensions

## No Symfony integration

- cache
- google-authenticator (abandoned)

All the checked ones have the AppKernel in place (so can benefit from this)
I would say we need to:

* Add a flag for this on the project configuration to mark the ones that can use the Symfony linter builds
* Fix doctrine-extensions to use AppKernel instead of Kernel

And later we can...

* Remove 3.x from BlockBundle (could it be done right now @VincentLanglet , or you prefer not to?)
* Remove 3.x from DoctrineMongoDBAdminBundle (I think this one is safe to do, never saw a request for bugfix on 3.x since the 4.0 release)
* Work on the rest of bundles / non bundles to provide the AppKernel

WDYT @VincentLanglet ?

templates/project/.github/workflows/symfony-lint.yaml.twig
```diff
@@ -0,0 +1,116 @@
+# DO NOT EDIT THIS FILE!
```
VincentLanglet, line 1: If I understand correctly, this file need to be added only if there is a test kernel.

VincentLanglet (changes requested):

templates/project/.github/workflows/symfony-lint.yaml.twig
jordisala1991: Yes, I have to make the needed changes.

VincentLanglet:
> Add a flag for this on the project configuration to mark the ones that can use the Symfony linter builds

It should be true by default, so our goal would be to have less and less project with `false`.

But instead of adding flag every times, I kinda wonder if we cannot auto detect some things.
For instance, instead of a phpstan flag, having a check to detect if phpstan is installed.
Here, the best would be to detect if the AppKernel class exists.


> Remove 3.x from BlockBundle (could it be done right now @VincentLanglet , or you prefer not to?)

I think too many stable/maintained version are relying on it IMHO

jordisala1991:
> But instead of adding flag every times, I kinda wonder if we cannot auto detect some things.
> For instance, instead of a phpstan flag, having a check to detect if phpstan is installed.
> Here, the best would be to detect if the AppKernel class exists.

I kinda like the reverse flag (true by default) because it allows to easily see what is missing to a bundle. Even though it is more config and more code...

templates/project/.github/workflows/symfony-lint.yaml.twig
jordisala1991: Fixed.

VincentLanglet (approved):

jordisala1991:
Some things that need to be fixed:

- Extensions package generate wrong namespace: the fix is to remove the Extensions part, just like we remove Sonata.
- MongoDB fails because the DB part gets lower cased, I guess the same is with ORM , not sure the best fix there.

Do you have some time for the fixes @VincentLanglet ? I can do them tomorrow otherwise.

VincentLanglet:
There are some fix to do @jordisala1991 

For instance the namespace of twig-extensions is just twig
https://github.com/sonata-project/twig-extensions/blob/1.x/tests/App/AppKernel.php

jordisala1991:
> There are some fix to do @jordisala1991
> 
> For instance the namespace of twig-extensions is just twig https://github.com/sonata-project/twig-extensions/blob/1.x/tests/App/AppKernel.php

https://github.com/sonata-project/SonataUserBundle/pull/1475
https://github.com/sonata-project/sonata-doctrine-extensions/pull/391

Those are the only two missing.

User-bundle seems to have some problem with FOSUser dependencies
Sonata-doctrine-extensions seems to not have all bundles needed to run all commands (probably because it does not have twig / yaml files) (haven't check)

We need to take a look at those.

Another thing to change is to add bin directory to export-ignore of gitattributes. Do you have time for that @VincentLanglet ?

VincentLanglet:
Sure https://github.com/sonata-project/dev-kit/pull/1856

VincentLanglet:
> Sonata-doctrine-extensions seems to not have all bundles needed to run all commands (probably because it does not have twig / yaml files) (haven't check)

Both Xliff and Twig check are failing because symfony/translations and twig are not installed.
Would it be possible to add a check in the github action ?

jordisala1991:
> > Sonata-doctrine-extensions seems to not have all bundles needed to run all commands (probably because it does not have twig / yaml files) (haven't check)
> 
> 
> 
> Both Xliff and Twig check are failing because symfony/translations and twig are not installed.
> 
> Would it be possible to add a check in the github action ?

Not sure how to do it tho, for now I disabled it for doctrine extensions. And im thinking to do the same for user bundle, we are not that far of having 5.0 version and we could enable it then.

PR #1845, opened 2022-01-03
Remove 3.x from block-bundle

jordisala1991:


jordisala1991:
It would be nice to have a way to create a release for an unmaintained branch. Like: this branch wont receive more dev kit updates but can receive fixes an we can create a release for it if needed.

VincentLanglet:
> It would be nice to have a way to create a release for an unmaintained branch. Like: this branch wont receive more dev kit updates but can receive fixes an we can create a release for it if needed.

Since the https://d5zda2diva-x6miu6vkqhzpi.eu.s5y.io/next-releases page cannot display all the release, it's maybe better to change the strategy by displaying a link for all the project stable branch.
This avoid to remember the project name used in the url, like
https://d5zda2diva-x6miu6vkqhzpi.eu.s5y.io/next-release/admin-bundle/4.x

Then it will be easy to change 4.x to 3.x if needed for instance.

VincentLanglet:
Also, we could drop 7.3 on most of the branches now https://www.php.net/supported-versions.php

jordisala1991:
> Also, we could drop 7.3 on most of the branches now https://www.php.net/supported-versions.php

Yes, we already drop from some projects.

VincentLanglet (approved):

core23 (approved):

PR #1847, opened 2022-01-03
Use composer-install v2, and use locked dependencies for dev-kit

jordisala1991:


jordisala1991:
When using locked deps it works (if we use composer.lock this is the mode we should use by default).

VincentLanglet (approved):

PR #1858, opened 2022-01-06
Bump phpstan to level 9

VincentLanglet:


src/Github/Api/Repositories.php
```diff
@@ -29,6 +29,9 @@ public function __construct(GithubClient $github)
         $this->github = $github;
     }
 
+    /**
+     * @return mixed[]
```
jordisala1991, line 33: I think we should be consistent here, you are using both: array<mixed> and mixed[] 

Why did you prefer array<mixed> on the other cases?

VincentLanglet: I generally prefer `array` because:
- You can add the keys more easily: from `array<string>` to `array<string, string>`.
- Phpstorm does not support `Foo[]` syntax when it's inside `array<>` or `array{}`, so it's better to write `array{foo: array<int>}` rather than `array{foo: int[]}`

But before using static analysis, only [] were supported so I still have habits.
Also, there is already lot of `PullRequest[]`, `string[]`, etc in the code, so it shouldn't be a blocker.
So the best would be to have a coding style rule, but it doesn't exists currently AFAIK.

jordisala1991 (approved):

PR #1890, opened 2022-01-21
Add phpunit_extensions extension to media-bundle

core23:
Fixes https://github.com/sonata-project/SonataMediaBundle/pull/2268


VincentLanglet:
It will add this on both 3.x and 4.x. I think we need to first install it on 3.x branch

PR #1893, opened 2022-01-27
Remove abandoned projects

VincentLanglet:


config/projects.yaml
```diff
@@ -295,20 +270,6 @@ media-bundle:
             variants:
                 symfony/symfony: ['4.4.*']
 
-notification-bundle:
```
core23, line 298: We should not remove this, because we have an active project that is still using this bundle:
https://github.com/sonata-project/SonataPageBundle/blob/3.x/composer.json#L36

core23 (changes requested):

config/projects.yaml
jordisala1991: 7 month ago was the last release, not sure if it is relevant to keep it here.

VincentLanglet: And it won't block us to make new releases if needed (or even add the project back here)

jordisala1991 (approved):

VincentLanglet:
Merging this will allow us to remove the phpstan/psalm option since all the maintained projet does have psalm and phpstan.

dmaicher (approved):

core23 (approved):
I'm not using the PageBundle anymore. So I can live with the removal

PR #1932, opened 2022-02-14
Add UserBundle 6.x branch to dev-kit

jordisala1991:


VincentLanglet:
It will consider the 5.x as the stable one then. Are we ok with this ?

jordisala1991:
To me, yes. The SonataUserBundle is almost ready

PR #1970, opened 2022-03-22
Initial rector integration

jordisala1991:


templates/project/.php-cs-fixer.dist.php
```diff
@@ -6,6 +6,17 @@
  * It's auto-generated by sonata-project/dev-kit package.
  */
 
+declare(strict_types=1);
```
VincentLanglet, line 9: In tests/bootstrap.php, we have the following order
image

- strict
- header
- uses
- DO NOT EDIT THIS FILE

In lot of other files, we have header before the `do not edit this file` message.

I would say we should always use the same order
- declare strict
- header
- Do not edit this file
- uses and php code

VincentLanglet (changes requested):

jordisala1991:
One thing I wanted to consider is how much dependencies rector has, and it is only dependant on php and phpstan, so it is a pretty safe addition to our composer.json (https://github.com/rectorphp/rector/blob/main/composer.json)

templates/project/.php-cs-fixer.dist.php
jordisala1991: Agree and fixed on every php file.

src/Domain/Value/Project.php
```diff
@@ -74,8 +75,8 @@ private function __construct(
         array $phpunitExtensions,
         bool $hasDocumentation,
         bool $hasTestKernel,
+        bool $rector,
         ?string $customDoctorRstWhitelistPart,
-        ?string $documentationBadgeSlug
```
OskarStark, line 78: Why remove this?

jordisala1991: Fixed.

src/Domain/Value/Project.php
```diff
@@ -123,6 +125,7 @@ public static function fromValues(string $name, array $config, Package $package)
             $phpunitExtensions,
             $config['has_documentation'],
             $config['has_test_kernel'],
+            $config['rector'],
             $config['custom_doctor_rst_whitelist_part'],
             $config['documentation_badge_slug'],
```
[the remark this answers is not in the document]
jordisala1991: Fixed.

VincentLanglet (approved):

jordisala1991:
I will wait before the merge to solve the admin bundle PR.

PR #1976, opened 2022-03-28
Use php-cs-fixer from the vendor

VincentLanglet:
Let's start to rely on the php-cs-fixer binary from the vendor instead of the dev one.

we'll add the dependency on each sonata project when using them. (solving a dev-kit PR or adding rector).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,18 +27,18 @@
         "league/commonmark": "^1.5",
         "nyholm/psr7": "^1.2",
         "oskarstark/env-var-extension": "^1.0",
-        "symfony/cache": "^5.3",
-        "symfony/console": "^5.3",
-        "symfony/dotenv": "^5.3",
+        "symfony/cache": "^5.4",
+        "symfony/console": "^5.4",
+        "symfony/dotenv": "^5.4",
         "symfony/flex": "^1.10",
-        "symfony/framework-bundle": "^5.3",
-        "symfony/http-client": "^5.3",
+        "symfony/framework-bundle": "^5.4",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -27,25 +27,26 @@
...
+        "symfony/twig-bundle": "^5.4",
+        "symfony/yaml": "^5.4",
         "twig/extra-bundle": "^3.1",
         "twig/markdown-extra": "^3.1",
         "twig/twig": "^2.12 || ^3.0",
         "webmozart/assert": "^1.9"
     },
     "require-dev": {
         "ergebnis/test-util": "^1.2",
+        "friendsofphp/php-cs-fixer": "^3.8",
```
jordisala1991, line 49: We should be careful with this one, it has a lot of dependencies, we should pick a minimum version that supports our whole version support (sf 4.4, 5.3, 6.0): https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/composer.json

Maybe on the dev-kit can stay with ^3.8, but we should pick ^3.4 for other Sonata Bundles (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/composer.json) so it is possible to install sf 4.4 on the prefer lowest range.

jordisala1991:
To me it is a bit problematic, also taking into account that php-cs-fixer is not specially fast adding support for things.

I think it might make sense to explore other options to include this php-cs-fixer, but not sure how (we agree that having to install it globally is wrong too).

jordisala1991 (approved):

PR #2007, opened 2022-04-28
add variant with BlockBundle 5 for AdminBundle 4

dmaicher:
See https://github.com/sonata-project/SonataAdminBundle/pull/7800#pullrequestreview-955210615

### DIFF
--- a/config/projects.yaml
+++ b/config/projects.yaml
@@ -11,6 +11,7 @@ admin-bundle:
             frontend: true
             variants:
                 symfony/symfony: ['4.4.*', '5.3.*', '5.4.*', '6.0.*']
+                sonata-project/block-bundle: ['5.x-dev as 5.0']
 
 block-bundle:
     rector: false
--- a/config/projects.yaml
+++ b/config/projects.yaml
@@ -11,7 +11,7 @@ admin-bundle:
             frontend: true
             variants:
                 symfony/symfony: ['4.4.*', '5.3.*', '5.4.*', '6.0.*']
-                sonata-project/block-bundle: ['5.x-dev as 5.0']
+                sonata-project/block-bundle: ['5.x-dev as 4.999']
 
 block-bundle:
     rector: false


config/projects.yaml
```diff
@@ -11,6 +11,7 @@ admin-bundle:
             frontend: true
             variants:
                 symfony/symfony: ['4.4.*', '5.3.*', '5.4.*', '6.0.*']
+                sonata-project/block-bundle: ['5.x-dev as 5.0']
```
dmaicher, line 14: is there any other way of allowing dev dependencies for a variant? Not so familiar with dev-kit

VincentLanglet: I don't see another solution. But I think it's dev-5.x.

But I think it could be better to write
```
sonata-project/block-bundle: ['dev-5.x as 4.999']
```

This avoid to add `^4 || ^5` in SonataAdmin composer.json which is kinda false since there is no 5.x version currently.

Also, I'm not sure this build should be required ; dunno how we could make this build optional
cc @jordisala1991

dmaicher: @VincentLanglet 

> But I think it's dev-5.x.

`5.x-dev` also works :wink: 

> This avoid to add ^4 || ^5 in SonataAdmin composer.json which is kinda false since there is no 5.x version currently.

I think the github action pipeline would do `composer require sonata-project/block-bundle:"5.x-dev as 5.0" or not? So this would alter the composer.json file anyway and no need for us to change it?

VincentLanglet: > I think the github action pipeline would do composer require sonata-project/block-bundle:"5.x-dev as 5.0" or not? So this would alter the composer.json file anyway and no need for us to change it?

This might work here for SonataAdmin/SonataBlock.

But let's say that `sonata-project/exporter` (or another dependency) require SonataBlock too.
- `composer require sonata-project/block-bundle:"5.x-dev as 5.0"` will give the error:
  - `sonata-project/exporter` doesn't support sonata-project/block-bundle 5.0.
- `composer require sonata-project/block-bundle:"5.x-dev as 4.999"` will work. 

I don't think any dependency will create an issue currently. But if you wanted to add variant with BlockBundle 5 on sonataMediaBundle, this would occur https://github.com/sonata-project/SonataMediaBundle/blob/4.x/composer.json#L86-L87 isn't it ?

dmaicher: Yes and I think its ok that those cases would not work. First we should then make sure that those dependencies are actually compatible with BlockBundle 5.

Currently we only really know that AdminBundle is compatible, or?

dmaicher: so @VincentLanglet @jordisala1991 how do we proceed? Should I change it? Or leave as is? I don't mind :blush:

VincentLanglet: I think `"5.x-dev as 4.999"` might be better WDYT @jordisala1991

jordisala1991: Yes, correct. It could also be like: "^5.0@dev as 4.999"

jordisala1991: Anyway, better to test it directly on admin-bundle and then make the modification needed here.

dmaicher: Yeah I added the variant manually and it passes: https://github.com/sonata-project/SonataAdminBundle/runs/6370000225?check_suite_focus=true

I can address the deprecations separately then

VincentLanglet (approved):

PR #2008, opened 2022-04-29
Add rector on block bundle

jordisala1991:
Related: https://github.com/sonata-project/SonataBlockBundle/pull/1057

### DIFF
--- a/templates/project/.github/workflows/qa.yaml.twig
+++ b/templates/project/.github/workflows/qa.yaml.twig
@@ -70,6 +70,7 @@ jobs:
             - name: Psalm
               run: vendor/bin/psalm --no-progress --show-info=false --stats --output-format=github --threads=$(nproc) --shepherd --php-version={{ branch.targetPhpVersion.toString }}
 {% if project.rector %}
+
     rector:
         name: Rector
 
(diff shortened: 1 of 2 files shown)


VincentLanglet (approved):

templates/project/.github/workflows/qa.yaml.twig
```diff
@@ -70,7 +70,6 @@ jobs:
             - name: Psalm
               run: vendor/bin/psalm --no-progress --show-info=false --stats --output-format=github --threads=$(nproc) --shepherd --php-version={{ branch.targetPhpVersion.toString }}
 {% if project.rector %}
-
```
VincentLanglet, line 73: We add a line between every step. Might be better to keep this line and add one before
```
psalm:
```

WDYT ?

jordisala1991: Yes we can do that instead.

templates/project/.github/workflows/qa.yaml.twig
```diff
@@ -70,7 +70,6 @@ jobs:
             - name: Psalm
               run: vendor/bin/psalm --no-progress --show-info=false --stats --output-format=github --threads=$(nproc) --shepherd --php-version={{ branch.targetPhpVersion.toString }}
 {% if project.rector %}
```
jordisala1991, line 72:
```suggestion
{% if project.rector %}

```

templates/project/.github/workflows/qa.yaml.twig
jordisala1991: Lets only enable rector for block bundle for now. Later we will fix this empty line issue

PR #2043, opened 2022-05-24
Little changes in gitignore and phpcs config

VincentLanglet:


templates/project/.gitignore.twig
```diff
@@ -2,6 +2,7 @@
 #
 # It's auto-generated by sonata-project/dev-kit package.
 
+.idea/
```
VincentLanglet, line 5: A recent contributor tried to commit those files.

Adding this could help some contributors and I dont see any drawback.

templates/project/.php-cs-fixer.dist.php
```diff
@@ -64,7 +64,6 @@
     'static_lambda' => true,
     'strict_comparison' => true,
     'strict_param' => true,
-    'types_spaces' => ['space' => 'single'],
```
VincentLanglet, line 67: Types_spaces is configured to 'none' by the Symfony standard and we tried to not override the default standard so far.

templates/project/.gitignore.twig
VincentLanglet: A recent contributor said he couldn't add it to a global ignore because he has to commit them in his work project. And that's a fair point.

If someone try to commit another IDE generated I won't be against adding it to the gitignore. IMO, it's the purpose of the gitignore to declare what we want to be committed and what we doesn't want to be ; we shouldnt rely on a gitignore global.
Also, I don't see any pain point for us to add this simple line.

VincentLanglet: WDYT @jordisala1991 ?

jordisala1991: I still think that opening the door for adding IDE files is a bad idea.

https://geekiam.io/why-you-and-your-team-should-have-a-global-git-ignore-file/
https://dev.to/oleksiyrudenko/please-do-not-pollute-projects-gitignore-3h8e

VincentLanglet: I disagree, but since there is no consensus, I removed this line.

franmomu (approved):

jordisala1991 (approved):

PR #2116, opened 2022-07-19
Allow to generate phpstan config with dev-kit

VincentLanglet:
We could consider to use something like this when all our stable branch are using level 8 phpstan. 
We can also add a support for `custom_phpstan.neon.dist` if needed. (We even can add the level in this custom config)

We might be able to do the same for psalm.

Related to suggestion from @eerison https://github.com/sonata-project/SonataPageBundle/pull/1469#issuecomment-1188704453 and @Hanmac https://github.com/sonata-project/SonataPageBundle/pull/1469#issuecomment-1189159517

WDYT @jordisala1991 ?


jordisala1991:
I really like the idea, the file exists is clever. I wanted to do this, but wasnt sure if it would work.

VincentLanglet:
This require more logic for SonataTranslation
https://github.com/sonata-project/SonataTranslationBundle/blob/3.x/phpstan.neon.dist

Also, current stable branch which are not using phpstan level 8:
- https://github.com/sonata-project/EntityAuditBundle/blob/1.x/phpstan.neon.dist
- https://github.com/sonata-project/SonataIntlBundle/blob/2.x/phpstan.neon.dist
- https://github.com/sonata-project/exporter/blob/2.x/phpstan.neon.dist
- PageBundle

src/Twig/Extension/FileSystemExtension.php
```diff
@@ -0,0 +1,43 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace App\Twig\Extension;
+
+use Symfony\Component\Filesystem\Filesystem;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+class FileSystemExtension extends AbstractExtension
```
OskarStark, line 20:
```suggestion
Final class FileSystemExtension extends AbstractExtension
```

VincentLanglet:
Now the AuditBundle use phpstan level 8 and now sonataIntl 3 is released, we only need PageBundle v4 to finish this.

VincentLanglet:
Let's try it

VincentLanglet:
This work perfectly.

We could try the same for psalm but:
- Some project are level 2 and some are level 3
- We ignore some specific things in the issueHandlers part

PR #2166, opened 2022-09-02
Update php-cs-fixer config

VincentLanglet:
Dunno if you want to split some "too long line"


jordisala1991 (approved):
For this config file I dont care a lot about long line tbh... problem for tomorrow, not today.

PR #2214, opened 2022-10-26
Test packages with PHP 8.2

franmomu:
POC: https://github.com/sonata-project/SonataAdminBundle/pull/7956 and https://github.com/sonata-project/SonataAdminBundle/actions/runs/3330179634/jobs/5508329137#step:3:14

We could start testing with PHP 8.2, should we add it to all of them (and fix the failing builds if any)? I may have some free time later today.

### DIFF
--- a/config/projects.yaml
+++ b/config/projects.yaml
@@ -2,12 +2,12 @@ admin-bundle:
     documentation_badge_slug: sonata-project-sonataadminbundle
     branches:
         5.x:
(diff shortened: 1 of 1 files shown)


VincentLanglet (approved):

VincentLanglet:
I think we can add it to all of them and see the failure, it will be quicker

VincentLanglet (approved):

PR #2348, opened 2023-03-05
drop sf 6.0 and 6.1 and use php 8.2 for everything except php-cs-fixer

jordisala1991:
For 6.x branch, the only release with maintenance is 6.2: https://symfony.com/releases

Needed for: https://github.com/sonata-project/SonataBlockBundle/pull/1139

I will later update all project maintained branches after we merge this one.

### DIFF
--- a/config/projects.yaml
+++ b/config/projects.yaml
@@ -3,16 +3,14 @@ admin-bundle:
     branches:
         5.x:
             php: ['7.4', '8.0', '8.1', '8.2']
-            target_php: '8.1'
             frontend: true
             variants:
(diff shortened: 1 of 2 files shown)


VincentLanglet (approved):

VincentLanglet:
Should we also drop PHP 7.4 ?

jordisala1991:
Php 7.4 should be dropped on november if we want to be strict with our policies (or on the next majors).

But the ideal would be to drop php up to 8.1 and sf up to 5.4.

PR #2367, opened 2023-03-23
Restrict packages to 4.4 correctly

jordisala1991:
This works around this issue: https://github.com/symfony/recipes/issues/1181

Proof: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1724

### DIFF
--- a/templates/project/.github/workflows/test.yaml.twig
+++ b/templates/project/.github/workflows/test.yaml.twig
@@ -82,6 +82,10 @@ jobs:
             - name: Add PHPUnit matcher
               run: echo "::add-matcher::{% verbatim %}${{ runner.tool_cache }}{% endverbatim %}/phpunit.json"
 
+            - name: Use older Flex endpoint for Symfony 4.4
+              if: matrix.symfony-require == '4.4.*'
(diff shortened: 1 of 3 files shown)


jordisala1991:
There are only a few repositories that do not maintain sf 4.4 , so we dont really need to add twig conditional to remove it from them. It can be done later if needed.

templates/project/.github/workflows/test.yaml.twig
```diff
@@ -82,6 +82,10 @@ jobs:
             - name: Add PHPUnit matcher
               run: echo "::add-matcher::{% verbatim %}${{ runner.tool_cache }}{% endverbatim %}/phpunit.json"
 
+            - name: Use older Flex endpoint for Symfony 4.4
+              if: matrix.symfony-require == '4.4.*'
+              run: composer config extra.symfony.endpoint https://raw.githubusercontent.com/symfony/recipes/c8d90298241f176a261dadc4814d7058415a4c1f/index.json
```
VincentLanglet, line 87: Should we add this conditionally to the fact we have symfony 4.4 in the variants ?
This way, the more we remove 4.4 support the more we won't have this in the yaml file

jordisala1991: Should be good now.

VincentLanglet (approved):

PR #2421, opened 2023-04-24
Create PR automatically

jordisala1991:
I think this could be a good first version of the automatic PR creation.

### DIFF
--- a/src/Command/ReleaseCommand.php
+++ b/src/Command/ReleaseCommand.php
@@ -17,20 +17,28 @@ use App\Action\DetermineNextRelease;
 use App\Action\Exception\NoPullRequestsMergedSinceLastRelease;
 use App\Config\Projects;
 use App\Domain\Value\Branch;
+use App\Domain\Value\NextRelease;
 use App\Domain\Value\Project;
 use App\Domain\Value\Stability;
+use App\Git\GitManipulator;
+use App\Github\Api\PullRequests;
 use App\Github\Domain\Value\CheckRuns;
 use App\Github\Domain\Value\CombinedStatus;
 use App\Github\Domain\Value\Label;
 use App\Github\Domain\Value\PullRequest;
+use Gitonomy\Git\Repository;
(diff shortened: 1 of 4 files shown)


jordisala1991:
Example PR: https://github.com/sonata-project/EntityAuditBundle/pull/566

jordisala1991:
Example run on github actions: https://github.com/sonata-project/dev-kit/actions/runs/4789783430

Example PR:
Example Release Draft:

jordisala1991:
This won't directly solve those PR that add a deprecated message like: This was deprecated on 4.x, and will be removed on 5.0.

You will still need to do that manually, but it is a nice step forward. With this I can release on the phone when the release is simple. You will only need to track the PR, merge when all green, and publish the release created as draft.

src/Command/ReleaseCommand.php
```diff
@@ -92,7 +111,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
         $this->io->getErrorStyle()->title($project->name());
 
-        return $this->renderNextRelease($project, $branch);
+        $nextRelease = $this->renderNextRelease($project, $branch);
+
+        if (null !== $nextRelease && $this->createPR($input, $output)) {
+            $this->prepareReleasePR($nextRelease);
+        }
+
+        return null !== $nextRelease ? 0 : 1;
```
VincentLanglet, line 120: We can use self::Success or self::Fail constant, I think something like this exist in Command instead of 0 or 1.

VincentLanglet (approved):
It could be helpful to know if we have to check for `@deprecated`/deprecations comment to update or not.

jordisala1991:
> It could be helpful to know if we have to check for `@deprecated`/deprecations comment to update or not.

Yes. That will be next change but I think it can be done on a next PR, there are no deprecation PRs lately, when the first one comes in, will improve this.

src/Command/ReleaseCommand.php
jordisala1991: Fixed.

PR #2423, opened 2023-04-25
Implement auto merge in github actions

jordisala1991:
Auto merge is not working since a month ago. Lets move it here and see logs better.


VincentLanglet:
> Auto merge is not working since a month ago. Lets move it here and see logs better.

It worked during some time, right after a new deploy. Not sure why ; but it will be better here indeed.

VincentLanglet (approved):

PR #2424, opened 2023-04-25
Update frontend

jordisala1991:
https://github.com/sonata-project/SonataAdminBundle/pull/7838


jordisala1991:
Even if we end up not adding Stimulus, this upgrade will be needed.

VincentLanglet (approved):

PR #2525, opened 2023-09-24
Update rector / php-cs-fixer config for test cases

core23:
Update rector / php-cs-fixer config for test cases.

POC: https://github.com/sonata-project/SonataAdminBundle/pull/8111

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -14,6 +14,9 @@ declare(strict_types=1);
 use Rector\Config\RectorConfig;
 use Rector\Php70\Rector\FunctionLike\ExceptionHandlerTypehintRector;
 use Rector\Php71\Rector\FuncCall\CountOnNullRector;
+use Rector\PHPUnit\Rector\Class_\AddSeeTestAnnotationRector;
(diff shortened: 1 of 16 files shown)


VincentLanglet (approved):

rector.php
```diff
@@ -24,12 +26,17 @@
 
     $rectorConfig->sets([
         LevelSetList::UP_TO_PHP_74,
+        PHPUnitSetList::PHPUNIT_90,
+        PHPUnitSetList::PHPUNIT_91,
```
OskarStark, line 30: Isn't there a `UP_TO_*` ?

core23: Good catch! 

https://github.com/rectorphp/rector-phpunit/blob/main/src/Set/PHPUnitLevelSetList.php

I'll update this later

VincentLanglet (approved):
The composer normalize is failing if you have time to fix it

core23:
Can we merge this? We might need some manual adjustments after dev-kit ran, because the phpunit data provider return type will be changed from `array` to `iterable`

VincentLanglet (approved):
it's ok to me

PR #2870, opened 2024-12-08
Update eslint rules

onEXHovia:
Disable rule for properties.

### DIFF
--- a/templates/project/.eslintrc.js
+++ b/templates/project/.eslintrc.js
@@ -39,5 +39,6 @@ module.exports = {
     ],
     'import/no-webpack-loader-syntax': 'off',
     'lines-between-class-members': { 'properties': 'off', 'methods': 'always' },
+    'arrow-body-style': 'off',
   },
 };
--- a/templates/project/.eslintrc.js
+++ b/templates/project/.eslintrc.js
@@ -38,5 +38,6 @@ module.exports = {
       2,
     ],
     'import/no-webpack-loader-syntax': 'off',
+    'lines-between-class-members': { 'properties': 'off', 'methods': 'always' },
   },
 };
--- a/templates/project/.eslintrc.js
+++ b/templates/project/.eslintrc.js
@@ -38,7 +38,8 @@ module.exports = {
       2,
     ],
     'import/no-webpack-loader-syntax': 'off',
-    'lines-between-class-members': { 'properties': 'off', 'methods': 'always' },
+    'lines-between-class-members': 'off',
+    'no-param-reassign': ['error', { 'props': false }],
     'arrow-body-style': 'off',
   },
 };


VincentLanglet:
Hi,

There is a 
```
/*
 * DO NOT EDIT THIS FILE!
 *
 * It's auto-generated by sonata-project/dev-kit package.
 */
 ```
 header in this file.
 
 This file is auto generated and dispatched from https://github.com/sonata-project/dev-kit/tree/master/templates/project
 
 So such change need to be done on the dev-kit repository.
And for such coding standard decision, it would be great to explain:
- Why disable `arrow-body-style`
- Why disable `lines-between-class-members` on properties

Cause I assume the `airbnb` standard was used to minimize such debate. It's not a big deal to have a new line between properties no ?

**Comment by onEXHovia** [comment]
The default configuration works, because code base use jquery.

I don't see the point in a rule that forces you to write
```js
export default class extends Controller {
  static targets = ['counter', 'field'];

  static outlets = ['filter'];
                                                        
  static classes = ['active'];
}
```

instead, right?
```js
export default class extends Controller {
  static targets = ['counter', 'field'];
  static outlets = ['filter'];               
  static classes = ['active'];
}
```

There should be a new line between methods, its good behavior. Same for the rule `arrow-body-style`
Disabling these rules will not create chaos in the code style, but they will remove annoying checks

onEXHovia:
Looking at checks https://github.com/sonata-project/SonataAdminBundle/actions/runs/12253259625/job/34181578909?pr=8230 we need to set up a rule `yoda`

because the package is supported by backend developers and code `null !== this.advancedTargets.find((advanced) => !advanced.hidden)` is normal for backend

VincentLanglet:
I'm ok for `lines-between-class-members`

I agree for `yoda` which need to be updated to be consistent with php-cs checks.

For `arrow-body-style` I'm not sure about the loss of forcing using arrow functions they have benefits I think.

VincentLanglet:
I dunno why I cannot give approval
image


I'll force the merge when the PR will be updated with all the needed rules.

onEXHovia:
`yoda` rule in next pr becauce affects another files
`no-param-reassign` for update property objects function params
allow behavior
https://github.com/sonata-project/SonataAdminBundle/pull/8230/files#diff-3a0b638bf958bb528c0a5318ef50022634d2199a23a204e6ce7e7ea3e44f9bd2R24
https://github.com/sonata-project/SonataAdminBundle/blob/4.x/webpack.config.js#L29

left the rule `arrow-body-style` for enable the use of braces around arrow function body
allow behavior
https://github.com/sonata-project/SonataAdminBundle/pull/8230/files#diff-9e471847050b0de785c04717ed47442f05b48c99ba2b4bf9347d5d79819e9c86R43

there is no reason to write
```js
  get enabledFields() {
    return this.fieldTargets.filter((field) => field.classList.contains(this.activeClass));
  }
```

instead
```js
  get enabledFields() {
    return this.fieldTargets.filter((field) => {
      return field.classList.contains(this.activeClass);
    });
  }
```

the rule doesn't affect anything, but it's always useful to make a block with return when the line is too long.

VincentLanglet:
Ok, and why you totally disable line between class members ? I thought it was ok to keep it for methods

onEXHovia:
> Ok, and why you totally disable line between class members ? I thought it was ok to keep it for methods

because admin bundle use 8.x branch eslint and localy failed to configure behavior, flexible settings is possible in 9.x far as I understand

VincentLanglet:
Ok, please order the rules alphabetically if possible.

And so far, I cannot merge this... because of github
image

VincentLanglet:
I successfully merged this manually, instead https://github.com/sonata-project/dev-kit/commit/5f40bd4dcecfadaf4f92650aa1f0860855364998

